PR #32, opened 2024-11-13
[Fix] Add <|eos|> as a stop token

Ubospica:
This PR adds <|eos|> as a stop token in the logic of detecting the stop token.


Ubospica:
#33

PR #40, opened 2024-11-16
[Performance] Rewrite the thread_safe cache.

DarkSharpness:
1. Add a cpp test on cache efficiency and correctness
2. Refactor the thread_safe_cache.

In the new thread_safe_cache, we maintain only 1 unified mapping instead of 2. In addition, in our implementation, we acquire the unique_lock less, which may bring a significant improvement in efficiency in some cases.

Here's the result for  `TEST(XGrammarParallelTest, CacheEfficiency)` on my machine `AMD EPYC 7R32`.

End to end time (preprocessing time `2s` not included):

- Old cache: `1152ms`
- New cache: `217ms`

gcc version: `13.2.0`

### DIFF
--- a/cpp/support/thread_safe_cache.h
+++ b/cpp/support/thread_safe_cache.h
@@ -7,7 +7,6 @@
 #define XGRAMMAR_SUPPORT_THREAD_SAFE_CACHE_H_
 
 #include <functional>
-#include <memory>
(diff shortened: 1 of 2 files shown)


Ubospica (approved):
Looks great. Thank @DarkSharpness for the contribution!

PR #61, opened 2024-11-18
[CUDA] JIT compilation for ApplyTokenBitmask kernel

MasterJH5574:
This PR introduces the JIT compilation for the CUDA kernel of ApplyTokenBitmask. The JIT compilation is enabled by the cuda-python package.

With JIT compilation, we can remove the AOT kernel compilation whcih introduces extra dependency when building the package.


yzh119 (approved):
cuda-python requires nvcc or nvrtc installation, nvcc is not available on pypi. You can consider adding [nvidia-cuda-nvrtc-cu12](https://pypi.org/project/nvidia-cuda-nvrtc-cu12/) to python package [requirements](https://github.com/mlc-ai/xgrammar/blob/33e3dc98563fd32a4c4286e7ef2ca85ed5357999/python/requirements.txt).

Also, cuda-python should be specified as a dependency as well.

python/xgrammar/cuda/apply_token_mask_inplace.py
```diff
@@ -0,0 +1,220 @@
...
+        batch_size, vocab_size = logits.shape
+    elif logits.ndim == 1:
+        batch_size = 1
+        (vocab_size,) = logits.shape
+    else:
+        raise ValueError(f"Invalid logits tensor shape {logits.shape}")
+    bitmask_size = (vocab_size + BITS_PER_BLOCK - 1) // BITS_PER_BLOCK
+
+    # Ensure that the tensors are contiguous in memory.
+    logits = logits.contiguous()
```
tqchen, line 182: need to check is_contiguous(), and raise an error if not. This is needed because logits is applied inplace

python/xgrammar/cuda/apply_token_mask_inplace.py
```diff
@@ -0,0 +1,220 @@
...
+    bitmask = bitmask.contiguous()
+
+    # Compile the kernel.
+    kernel = KernelStore.compile(logits.device.index)
+    # Setup kernel launching arguments.
+    grid_dims = (batch_size * bitmask_size + THREADS_PER_BLOCK - 1) // THREADS_PER_BLOCK, 1, 1
+    block_dims = THREADS_PER_BLOCK, 1, 1
+    shared_mem_bytes = 0
+    stream = cuda.CU_STREAM_LEGACY
```
tqchen, line 191: https://pytorch.org/docs/stable/generated/torch.cuda.current_stream.html

python/xgrammar/cuda/apply_token_mask_inplace.py
```diff
@@ -0,0 +1,220 @@
...
+        module = checkCudaErrors(cuda.cuModuleLoadData(np.char.array(data)))
+        func = checkCudaErrors(cuda.cuModuleGetFunction(module, b"ApplyTokenBitmaskInplaceKernel"))
+        cls._module = module
+        cls._func = func
+        # Return the compiled kernel.
+        return func
+
+
+def apply_token_bitmask_inplace(logits: torch.Tensor, bitmask: torch.Tensor):
+    if cuda is None or cudart is None or nvrtc is None:
```
tqchen, line 168: cjecl deboce type of logits and bitmask and raise an error

tqchen (approved):

tqchen:
I left some comments that should be done ine followup PR , cc @CharlieFRuan @Ubospica @MasterJH5574

Handling dependency is tricky, given there can be cuda and non-cuda variants. Here is what we can do here

Add clear instruction to install cuda related dependency in the install guide

```bash

## Install from PyPI

.. code::
   
    pip install xgrammar


## CUDA Dependencies

Please also install these extra dependencies to enable CUDA support for masking

.. code::
 
    pip install cuda-python nvidia-cuda-nvrtc-cu12 

```

Additionally, have a clear error messages when the dependency is not present

```python

cuda_dep_msg = """
Cannot find cuda-python, please install via https://xgrammar.docs.ai/docs/install.html#cuda-dependencies
"""
```

MasterJH5574:
Followed up in #64 and resolved the comments.

PR #75, opened 2024-11-22
[Compiler] Special case max_thread=1 to not use ThreadPool or mutex

CharlieFRuan:
This PR bypasses using `ThreadPool` or `std::mutex` when `max_threads=1` when compiling grammar. This is motivated by how WebAssembly does not have good support for multithreading.

Unrelated to this, fix a bug in CPU bitmasking.

Tested with `pytest tests/`.

### DIFF
--- a/cpp/compiler.cc
+++ b/cpp/compiler.cc
@@ -238,12 +238,12 @@ CompiledGrammar MultiThreadCompileGrammar(
   // TODO(Charlie): Figure out how to support ThreadPool and std::mutex in WebAssembly.
   // Only declare ThreadPool and mutex if max_threads > 1, so when max_threads = 1, we do
   // not need ThreadPool or std::mutex, which throws error in runtime in WebAssembly.
-  std::unique_ptr<ThreadPool> thread_pool;
-  std::unique_ptr<std::mutex> adaptive_token_mask_cache_mutex;
+  std::optional<ThreadPool> thread_pool;
+  std::optional<std::mutex> adaptive_token_mask_cache_mutex;
 
   if (max_threads > 1) {
-    thread_pool = std::make_unique<ThreadPool>(max_threads);
-    adaptive_token_mask_cache_mutex = std::make_unique<std::mutex>();
+    thread_pool.emplace(max_threads);
+    adaptive_token_mask_cache_mutex.emplace();
   }
 
   auto root_rule_id = grammar->GetRootRuleId();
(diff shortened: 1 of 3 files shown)


cpp/compiler.cc
```diff
@@ -254,16 +262,22 @@ CompiledGrammar MultiThreadCompileGrammar(
...
           auto add_adaptive_token_mask = [&](const RulePosition& rule_position) {
             auto grammar_matcher = GrammarMatcherForCompiler(grammar, rule_position);
             auto cur_adaptive_token_mask_cache = grammar_matcher.GetAdaptiveTokenMask(
                 tokenizer_info.GetVocabSize(),
                 tokenizer_info.GetSortedDecodedVocab(),
                 rule_id != root_rule_id
             );
-            {
-              std::lock_guard<std::mutex> lock(adaptive_token_mask_cache_mutex);
+            if (mutex_ptr) {
```
Ubospica: use max_threads > 1 as condition for simplicity

cpp/compiler.cc
```diff
@@ -254,16 +262,22 @@ CompiledGrammar MultiThreadCompileGrammar(
         if (element.type == RuleExprType::kRuleRef) {
           continue;
         }
-        thread_pool.Execute([&, rule_id, sequence_id, element_id, element]() {
+        // Define the per-element processing logic for code reuse between
+        // using thread_pool and not using thread_pool
+        auto process_element = [&, rule_id, sequence_id, element_id, element](std::mutex* mutex_ptr
```
Ubospica: capture adaptive_token_mask_cache_mutex by reference instead of passing it as a ptr

cpp/compiler.cc
CharlieFRuan: Addressed

cpp/compiler.cc
CharlieFRuan: Addressed

cpp/compiler.cc
```diff
@@ -235,8 +235,16 @@ CompiledGrammar MultiThreadCompileGrammar(
   // 1. All character class or character class star (with last_utf8_bytes=0, 1, 2, 3)
   // 2. All byte strings (with element_in_string=0, 1, 2, ...)
 
-  ThreadPool thread_pool(max_threads);
-  std::mutex adaptive_token_mask_cache_mutex;
+  // TODO(Charlie): Figure out how to support ThreadPool and std::mutex in WebAssembly.
+  // Only declare ThreadPool and mutex if max_threads > 1, so when max_threads = 1, we do
```
tqchen, line 239: we don't have to do that as threading in webassembly is tricky and needs webworkers

PR #100, opened 2024-11-26
add triton mask kernel implementations

yzh119:
This PR adds triton implementation of the mask kernels because triton is easier and more friendly to maintain.

This is just a proof of concept, and I haven't tuned performance yet, leave it for future work.

cc @Ubospica @MasterJH5574


yzh119:
As we discussed earlier, one optimization is to remove the `tl.load(logits_ptr ...`, but only store inf to certain positions:
```python
        tl.store(logits_ptr + batch_id * vocab_size + offsets, -float("inf", vocab_mask & bitmask)
```

PR #109, opened 2024-12-03
[Fix] Fix local Python package installation

Ubospica:
This PR fixes the problem in local installation of python package. It fixes the wrong path of `version.py` in `setup.py`.

Co-authored-by: zanderjiang <zander.jiang@gmail.com>


Ubospica:
cc @zanderjiang

PR #116, opened 2024-12-05
sentence piece tokenizer support for TokenizerInfo

zanderjiang:


python/xgrammar/tokenizer_info.py
```diff
@@ -196,8 +224,65 @@ def from_huggingface(
...
+        
+            # fill in base vocabulary from sp_model
+            for i in range(sp_model.get_piece_size()):
+                piece = sp_model.id_to_piece(i)
+                if i < vocab_size:
+                    encoded_vocab[i] = piece
+        
+            vocab_dict = tokenizer.get_vocab()
+        
+            # fill in any special tokens from vocab_dict
```
Ubospica, line 252: Why don't we just construct encoded_vocab from vocab_dict? Also, we can use list comprehension, which is more efficient.

python/xgrammar/tokenizer_info.py
```diff
@@ -196,8 +224,65 @@ def from_huggingface(
...
+            for token, idx in vocab_dict.items():
+                if idx < vocab_size:
+                    encoded_vocab[idx] = token
+
+            # detect vocab_type of tokenizer
+            byte_level_tokens = ["Ġ", "Ċ", "ĉ", "Ā"]
+            if any("<0x" in token for token in vocab_dict):
```
Ubospica: Just checking one token (e.g. <0x0a>) is ok. 
```suggestion
            if "<0x0a>" in vocab_dict:
```
That should be more efficient

Ubospica (commented):
It looks great overall. Thanks for the contribution!

PR #143, opened 2025-01-08
parser for function calling

zanderjiang:


cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+std::vector<std::pair<std::string, std::unordered_map<std::string, std::string>>> parse_message(
+    const std::string& message_body
+) {
+  std::vector<std::pair<std::string, std::unordered_map<std::string, std::string>>> tool_calls;
+  size_t i = 0;
+  const size_t length = message_body.length();
+
+  while (i < length) {
+    // <function> tags
+    if (i + 10 < length && message_body.substr(i, 10) == "<function=") {
```
Ubospica: Use std::string_view(s).substr(i, 10) to avoid allocating new memory

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+#include <unordered_map>
+
+#include "picojson.h"
+
+namespace xgrammar {
+
+/******************* Parser *******************/
+
+std::vector<std::pair<std::string, std::unordered_map<std::string, std::string>>> parse_message(
+    const std::string& message_body
```
Ubospica: add a param bool ignore_error, default false
if true, skip all error messages (just like now). 

Otherwise, raise ValueError (ie. XGRAMMAR_CHECK or XGRAMMAR_LOG(FATAL)) at all error places (ie the "continue" places in the current code)

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+      std::string params_str = message_body.substr(params_start, params_end - params_start);
+      size_t pos = 0;
+      while ((pos = params_str.find('\'', pos)) != std::string::npos) {
+        params_str.replace(pos, 1, "\"");
+        pos++;
+      }
+
+      picojson::value v;
+      std::string err = picojson::parse(v, params_str);
+      if (!err.empty()) {
```
Ubospica: if error is not empty or v is not an object

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+        i++;
+        continue;
+      }
+
+      if (v.is<picojson::object>()) {
```
Ubospica: we shall check error first, instead of putting long piece of code in a nested scope. So
```
if (!...) { raise error; }
```
is better than
```
if (...) {
    ...
} else {
    raise error;
}

File: cpp/function_calling.cc [hunk 5]
```diff
@@ -0,0 +1,124 @@
...
+          if (pair.second.is<std::string>()) {
+            params[pair.first] = pair.second.get<std::string>();
+          }
+        }
+        tool_calls.emplace_back(function_name, params);
+      }
+
+      i = params_end + 11;
+    }
+    // json format
```
Ubospica: Add more detailed comments; provide an example

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+
+std::vector<std::pair<std::string, std::unordered_map<std::string, std::string>>> parse_message(
+    const std::string& message_body
+) {
+  std::vector<std::pair<std::string, std::unordered_map<std::string, std::string>>> tool_calls;
+  size_t i = 0;
+  const size_t length = message_body.length();
+
+  while (i < length) {
+    // <function> tags
```
Ubospica: Add more detailed comments; provide an example

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+            params[pair.first] = pair.second.get<std::string>();
+          }
+        }
+        tool_calls.emplace_back(function_name, params);
+      }
+
+      i = params_end + 11;
+    }
+    // json format
+    else if (i < length && message_body[i] == '{') {
```
Ubospica:
```
} else if () {
  // Explain...
}
```

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+            for (const auto& pair : params_obj) {
+              if (pair.second.is<std::string>()) {
+                params[pair.first] = pair.second.get<std::string>();
+              }
+            }
+            tool_calls.emplace_back(name, params);
+          }
+        }
+        i = j;
+      } else {
```
Ubospica: Always put short cases first and avoid nesting for long cases if possible

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+          i = j;
+          continue;
+        }
+
+        if (v.is<picojson::object>()) {
+          const picojson::object& obj = v.get<picojson::object>();
+          auto name_it = obj.find("name");
+          auto params_it = obj.find("parameters");
+
+          if (name_it != obj.end() && params_it != obj.end() && name_it->second.is<std::string>() &&
```
Ubospica: if not, report error or continue

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+      }
+
+      i = params_end + 11;
+    }
+    // json format
+    else if (i < length && message_body[i] == '{') {
+      int bracket_count = 1;
+      size_t j = i + 1;
+
+      while (j < length && bracket_count > 0) {
```
Ubospica: prefer for loop than while

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+          auto params_it = obj.find("parameters");
+
+          if (name_it != obj.end() && params_it != obj.end() && name_it->second.is<std::string>() &&
+              params_it->second.is<picojson::object>()) {
+            std::string name = name_it->second.get<std::string>();
+            const picojson::object& params_obj = params_it->second.get<picojson::object>();
+
+            std::unordered_map<std::string, std::string> params;
+            for (const auto& pair : params_obj) {
+              if (pair.second.is<std::string>()) {
```
Ubospica: also, check and report error

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+
+namespace xgrammar {
+
+/******************* Parser *******************/
+
+std::vector<std::pair<std::string, std::unordered_map<std::string, std::string>>> parse_message(
+    const std::string& message_body
+) {
+  std::vector<std::pair<std::string, std::unordered_map<std::string, std::string>>> tool_calls;
+  size_t i = 0;
```
Ubospica: use more meaningful var names, eg input_it, instead of i and j

cpp/function_calling.cc
```diff
@@ -0,0 +1,124 @@
...
+            std::string name = name_it->second.get<std::string>();
+            const picojson::object& params_obj = params_it->second.get<picojson::object>();
+
+            std::unordered_map<std::string, std::string> params;
+            for (const auto& pair : params_obj) {
+              if (pair.second.is<std::string>()) {
+                params[pair.first] = pair.second.get<std::string>();
+              }
+            }
+            tool_calls.emplace_back(name, params);
```
Ubospica:
```suggestion
            tool_calls.emplace_back(name, std::move(params));
```

Ubospica (commented):
Thx for the contribution, please finish the Python part and tests

cpp/pybind/pybind.cc
```diff
@@ -86,7 +86,8 @@ PYBIND11_MODULE(xgrammar_bindings, m) {
               bool>(&JSONSchemaToEBNF)
       )
       .def("_regex_to_ebnf", &RegexToEBNF)
-      .def("_get_masked_tokens_from_bitmask", &Matcher_DebugGetMaskedTokensFromBitmask);
+      .def("_get_masked_tokens_from_bitmask", &Matcher_DebugGetMaskedTokensFromBitmask)
+      .def("_parse_message", &parse_message, py::arg("input"), py::arg("ignore_error") = false);
```
Ubospica: Dont need to specify args. Just set default args in python bindings

PR #153, opened 2025-01-14
[3rdparty] Avoid macro redefinition in picojson

MasterJH5574:
This PR adds a guard to avoid the macro redifinition of `__STDC_FORMAT_MACROS` in 3rdparty picojson.


MasterJH5574:
cc @Ubospica

Ubospica:
Looks great. Thanks! @MasterJH5574

PR #162, opened 2025-01-21
[Feature] Structural tag

Ubospica:
This PR 
- Supports the parsing and mask generation algorithm for TagDispatch in GrammarMatcherBase and GrammarMatcher
- Refactors GrammarMatcherBase and GrammarMatcher along the way
- Designs the CSRArray and FSM data structures which are useful in later architecture
- Supports the structural tag feature with the above backend support.

Description of the structural tag:

The structural tag handles the dispatching
of different grammars based on the tags and triggers: it initially allows any output,
until a trigger is encountered, then dispatch to the corresponding tag; when the end tag
is encountered, the grammar will allow any following output, until the next trigger is
encountered.

The tags parameter is used to specify the output pattern. It is especially useful for LLM
function calling, where the pattern is:
<function=func_name>{"arg1": ..., "arg2": ...}</function>.
This pattern consists of three parts: a start tag (<function=func_name>), a parameter list
according to some schema ({"arg1": ..., "arg2": ...}), and an end tag (</function>). This
pattern can be described in a StructuralTagItem with a start tag, a schema, and an end tag.
The structural tag is able to handle multiple such patterns by passing them into multiple
tags.

The triggers parameter is used to trigger the dispatching of different grammars. The trigger
should be a prefix of a provided start tag. When the trigger is encountered, the
corresponding tag should be used to constrain the following output. There can be multiple
tags matching the same trigger. Then if the trigger is encountered, the following output
should match one of the tags. For example, in function calling, the triggers can be
["<function="]. Then if "<function=" is encountered, the following output must match one
of the tags (e.g. `<function=get_weather>{"city": "Beijing"}</function>`).

The corrrespondence of tags and triggers is automatically determined: all tags with the
same trigger will be grouped together. User should make sure any trigger is not a prefix
of another trigger: then the corrrespondence of tags and triggers will be ambiguous.

To use this grammar in grammar-guided generation, the GrammarMatcher constructed from
structural tag will generate a mask for each token. When the trigger is not encountered,
the mask will likely be all-1 and not have to be used (fill_next_token_bitmask returns
False, meaning no token is masked). When a trigger is encountered, the mask should be
enforced (fill_next_token_bitmask will return True, meaning some token is masked) to the
output logits.

The benefit of this method is the token boundary between tags and triggers is automatically
handled. The user does not need to worry about the token boundary.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@
 #     pre-commit autoupdate
 #
 # See https://github.com/pre-commit/pre-commit
-# Note the pre-commit hooks shoule only be used for formatting, but not for linting.
+# Note the pre-commit hooks should only be used for formatting, but not for linting.
 # For linting consider using CI.
 repos:
   # Standard hooks
--- a/cpp/compiled_grammar_data_structure.h
+++ b/cpp/compiled_grammar_data_structure.h
@@ -66,6 +66,11 @@ struct AdaptiveTokenMask {
       const std::vector<int32_t>& rejected_indices,
       const std::vector<int32_t>& uncertain_indices
   );
+
+  static std::string PrintTokenByIds(
+      const std::vector<int32_t>& token_ids, const TokenizerInfo& tokenizer_info, int max_print_num
+  );
+  std::string Print(const TokenizerInfo& tokenizer_info) const;
 };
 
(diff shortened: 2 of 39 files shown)


Ubospica:
cc @tqchen @zanderjiang

PR #164, opened 2025-01-23
[Fix] Rollback safely when token acceptance fails

benchislett:
In `GrammarMatcher::Impl::AcceptToken`, if a char is not accepted, the matcher state is left broken because the partially-advanced token state remains in the matcher.

The below function, `GrammarMatcher::Impl::_DebugAcceptString`, handles this case correctly. I'm not sure why it isn't present in AcceptToken.

If this behaviour is not deliberate, please accept this fix so that AcceptToken can handle rejections gracefully.


Ubospica:
Hi @benchislett, goot catch and thanks for the contribution! Could you also add some test cases in `tests/python/test_grammar_matcher.py`

benchislett:
@Ubospica done! Any other feedback is welcome.

PR #170, opened 2025-01-26
fix(test): fix the concurrency issue in #168 

### LINKED ISSUE #168: Flaky CPP test - CacheCorrectness
I'm getting around a 20% failure rate on this test with 4 cores.
`xgrammar_test --gtest_filter=XGrammarParallelTest.CacheCorrectness`

A mutex is needed when accessing the lifespan hook map here https://github.com/mlc-ai/xgrammar/blob/5ee42566168049d32c2ef4e90374f03ed80722d8/tests/cpp/test_thread_safe_cache.cc#L84

The reason being is that many threads return a copy of the cached TestObject. The copy operation is not under the protection of the cache lock so the lifespan map is updated unsafely.

DarkSharpness:
This PR aims to fix #168 

The `unordered_map` in `LifeSpanHook` is not thread safe, so we need to protect it with a lock.

### DIFF
--- a/tests/cpp/test_thread_safe_cache.cc
+++ b/tests/cpp/test_thread_safe_cache.cc
@@ -5,6 +5,7 @@
 #include <chrono>
 #include <cstddef>
 #include <future>
+#include <mutex>
 #include <string>
 #include <thread>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


chris-monardo (approved):

Ubospica:
It looks great! Thanks @DarkSharpness

PR #171, opened 2025-01-26
fix(fsm): fix error of dangling-reference in gcc-13

DarkSharpness:
When compiling with `gcc-13.2.0`, we encountered a false-positive "dangling-reference" warning.

To mitigate this issue, we now construct the object outside the inner loop to prevent the false positive.

For more details on this gcc bug, refer to [this article](https://trofi.github.io/posts/264-gcc-s-new-Wdangling-reference-warning.html).


Ubospica:
Thanks for fixing the issue!

PR #177, opened 2025-01-31
[Resource] Add XGrammar Slides

Ubospica:
This PR adds a slide introducing XGrammar


Ubospica:
This PR should be reverted because we do not wish to upload a large file to the repo. To avoid adding the size of the `.git` folder, the commit is deleted and the main branch is force-pushed, instead of just reverting this PR.

PR #180, opened 2025-02-03
test lazy character class

qeternity:
There is an issue with the compiler and lazy character classes.

The test I have added will raise the following exception:
`RuntimeError: [12:25:32] /workspace/cpp/grammar_parser.cc:78: EBNF parse error at line 1, column 16: Expect element`


Ubospica:
Hi @qeternity, thanks for the contribution! The non-greedy modifier `?` should indeed be supported. We have supported it in #184. We can merge this PR given the test can pass now.

PR #185, opened 2025-02-06
[Feature] Support Regex for GrammarCompiler

Ubospica:
This PR supports regex for grammar compiler, finalizing the support for regex in xgrammar.


qeternity:
Thanks for the @Ubospica. Could we get a minor release containing this and some of the regex fixes?

Would like to update SGLang.

Ubospica:
Hi @qeternity , v0.1.12 is just released. I hope this would fit into SGLang's workflow!

PR #186, opened 2025-02-07
[Feature] Support Optimized CUDA Token Bitmask Kernel

syuoni:
This PR supports CUDA-implemented token bitmask kernel. It shows better perf than the Triton-implemented one, especially on large batch sizes.

Also, the PR provides `examples/benchmark/bench_apply_token_bitmask_inplace.py` for benchmarking. ~~The perf results are:~~

~~| GPU            | Batch size | Vocab size | Triton (μs)  | CUDA (μs) | Speedup ratio  |
|:--------------:|-----------:|-----------:|-------------:|----------:|---------------:|
| H100 80GB HBM3 | 16         | 128k       | 75.4         | 6.5       | 11.65x         |
|                | 128        | 128k       | 79.0         | 52.4      | 1.51x          |
|                | 2048       | 128k       | 1048.5       | 714.5     | 1.47x          |
| A100 SXM4 80GB | 16         | 128k       | 137.1        | 9.6       | 14.30x         |
|                | 128        | 128k       | 140.3        | 88.0      | 1.59x          |
|                | 2048       | 128k       | 1439.9       | 1293.2    | 1.11x          |~~

Please checkout `examples/benchmark/README.md` for the benchmarking results.

----
Notes:
* Users can switch to the Triton version by setting env `XGRAMMAR_TOKEN_BITMASK_TRITON=1`.
* The PR is adapted from https://github.com/NVIDIA/TensorRT-LLM/blob/v0.17.0/cpp/tensorrt_llm/kernels/logitsBitmask.cu


examples/benchmark/bench_apply_token_bitmask_inplace.py
```diff
@@ -0,0 +1,66 @@
...
+
+    start = time.perf_counter()
+    for i in range(args.num_iters):
+        run()
+    torch.cuda.synchronize()
+    exec_time = time.perf_counter() - start
+    exec_time = (exec_time / args.num_iters) * 10**6
+
+    print(f"Kernel: {args.kernel}")
+    print(f"Kernel execution time (us): {exec_time:.4f}")
```
Ubospica: Could you put these tests into `tests/python/test_token_bitmask_operations.py`? Also, can you consider the mask to be all full, half full, or almost all empty during benchmarking? We found in testing that this has a major impact on the kernel execution time.

It should be easy because we have provided all utilities for this in `test_apply_token_bitmask_inplace_large()`. You can just fit into the workflow of that function.

Ubospica (commented):
Thanks for the contribution @syuoni ! The enhancement of kernels is definitely useful. Could you please modify the benchmark accordingly? Once that’s done, we can proceed with merging this PR.

examples/benchmark/bench_apply_token_bitmask_inplace.py
syuoni: Seems when `masked_cnt` is small, the triton kernel skips a lot of memory access, which results higher perf. Let me check if the CUDA kernel can adopt this pattern.

In a realistic scenario, what is the distribution of the bitmask over the three patterns?
* full 
* half full
* almost all empty

BTW, does "full" mean ALL logits filled with -inf or NO logits filled with -inf?

Ubospica: I think it should be possible to take advantage of Triton kernels to optimize CUDA code.

When I say “all full,” I mean the case where all values are -1 (all-true), which corresponds to the case where mask_cnt is very small - just a terminology issue.

In real cases, I believe both the “full” and “almost all empty” cases are quite common. For example, in the case of JSON, almost all tokens are valid inside a string, which corresponds to the “full” case. On the other hand, outside a string, only tokens that conform to the syntax are valid, which corresponds to the “all-empty” case.

examples/benchmark/README.md
```diff
@@ -15,3 +15,50 @@ lm-format-enforcer                0.10.6
...
+|                |          8 |       128k |       127k |        13.45 |     11.27 |   1.19x |
+|                |         64 |       128k |         1k |        22.95 |     25.57 |   0.90x |
+|                |         64 |       128k |        64k |        58.52 |     56.47 |   1.04x |
+|                |         64 |       128k |       127k |        44.83 |     39.29 |   1.14x |
+|                |        512 |       128k |         1k |       132.92 |    108.60 |   1.22x |
+|                |        512 |       128k |        64k |       362.08 |    349.54 |   1.04x |
+|                |        512 |       128k |       127k |       306.75 |    233.20 |   1.32x |
+|                |       4096 |       128k |         1k |       955.99 |    777.94 |   1.23x |
+|                |       4096 |       128k |        64k |      2756.63 |   2707.57 |   1.02x |
+|                |       4096 |       128k |       127k |      2472.82 |   1782.41 |   1.39x |
```
syuoni, line 64: Hi @Ubospica ,

Please checkout the updated benchmarking results (using `triton.testing.do_bench`). The masked cnt of 1k, 64k and 127k corresponds to (almost) all full, half full and almost empty cases.

Refreshed benchmarking results:
* CUDA kernel shows similar latencies to Triton kernel on small batch sizes
* CUDA kernel offers significant perf gains on large batch sizes (on most cases).

python/xgrammar/kernels/apply_token_bitmask_inplace_triton.py
```diff
@@ -21,7 +21,7 @@ def apply_token_bitmask_inplace_kernel(
     for work_id in tl.range(pid, num_rows * num_blocks, NUM_SMS):
         block_offset = (work_id % num_blocks) * BLOCK_SIZE
         row_id = work_id // num_blocks
-        batch_id = tl.load(indices_ptr + row_id)
+        batch_id = row_id if indices_ptr is None else tl.load(indices_ptr + row_id)
```
syuoni, line 24: This is an optimization for the Triton kernel.

For contiguous logits tensor, we can pass `indices_ptr=None` to the kernel, and thus the kernel can skip loading `indices_ptr`. This saves global memory access. Surprisingly, this can causes up to 30% perf differences.

python/xgrammar/testing.py
```diff
@@ -158,6 +158,34 @@ def _get_masked_tokens_from_bitmask(
     )
 
 
+def _bool_mask_to_bitmask(bool_mask: torch.Tensor) -> torch.Tensor:
```
syuoni, line 161: I moved this `_bool_mask_to_bitmask` function from `tests/python/test_token_bitmask_operations.py`. So that this function can be shared to the benchmarking script.

tests/python/test_token_bitmask_operations.py
```diff
@@ -67,67 +57,86 @@ def test_apply_token_bitmask_inplace(is_cuda: bool):
...
+    if masked_cnt >= vocab_size:
+        bool_mask = torch.zeros(batch_size, vocab_size, dtype=torch.bool)
+    else:
+        bool_mask = torch.ones(batch_size, vocab_size, dtype=torch.bool)
+        if masked_cnt > 0:
+            masked_positions = torch.stack(
+                [torch.randperm(vocab_size)[:masked_cnt] for _ in range(batch_size)]
+            )
+            bool_mask.scatter_(1, masked_positions, False)
+            assert (bool_mask.sum(dim=-1) + masked_cnt == vocab_size).all().item()
```
syuoni, line 99: I changed the approach to generate `bool_mask`. 

The original approach does not ensure there is `masked_cnt` positions being masked, because there may be repeated values in `masked_positions`.

python/xgrammar/kernels/apply_token_bitmask_inplace_triton.py
Ubospica: That is quite interesting. Thanks for pointing that out!

Ubospica:
This PR looks great to me. The efficiency of the applymask kernel is critical for end-to-end LLM serving. Thanks @syuoni !

Ubospica:
Hi @syuoni, I would like to ask about these lines in this PR:
https://github.com/mlc-ai/xgrammar/blob/main/python/xgrammar/kernels/apply_token_bitmask_inplace_cuda.py#L61-L67

> @torch.library.register_fake("xgrammar::apply_token_bitmask_inplace_cuda")
> def _(
>     logits: torch.Tensor,
>     bitmask: torch.Tensor,
>     indices: Optional[torch.Tensor] = None,
> ) -> None:
>     pass

Do you think this FakeTensor op registration has any purpose? 

The `torch.library.register_fake` is supported in PyTorch 2.4, but we want to maintain compatibility with low PyTorch versions. If this function is not necessary for your application, maybe I can remove it; otherwise, I will add a check for the existence of it first.

syuoni:
Hi @Ubospica ,

Removing this fake registration function will not break `torch.compile` with `fullgraph=True`. Try below code:

```python
import torch
import xgrammar


@torch.compile(fullgraph=True)
def func(logits, bitmask):
    xgrammar.kernels.apply_token_bitmask_inplace_cuda(logits, bitmask)


if __name__ == "__main__":
    logits = torch.randn(128000, device=0)
    bitmask = torch.randn(128000//32, device=0).view(torch.int32)
    func(logits, bitmask)
```

It raises error:
```bash
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/torch/_subclasses/fake_tensor.py", line 2378, in _dispatch_impl
    r = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/_ops.py", line 723, in __call__
    return self._op(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
NotImplementedError: xgrammar::apply_token_bitmask_inplace_cuda: attempted to run this operator with Meta tensors, but there was no fake impl or Meta kernel registered. You may have run into this message while using an operator with PT2 compilation APIs (torch.compile/torch.export); in order to use this operator with those APIs you'll need to add a fake impl. Please see the following for next steps:  https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html
```

If users don't use `torch.compile` with `fullgraph=True` (or similarly, `torch.export`), removing the fake registration function is OK.

Seems [torch.library.impl_abstract](https://pytorch.org/docs/stable/library.html#torch.library.impl_abstract) is the name for `torch.library.register_fake` in lower PyTorch version.

Ubospica:
@syuoni Thanks for the reply. That makes sense.

`torch.library.impl_abstract` looks good as well but is not supported till torch 2.2. Now #206 will check if the property exists and call it. I think this should be fine as well (if you still see any issues, please let me know ：) ).

syuoni:
https://github.com/mlc-ai/xgrammar/pull/206 looks good to me. Thanks!

PR #190, opened 2025-02-12
Modernize build to support sdists

zbowling:
Replace the two setup setuptools based build process with scikit_build_core.build to call cmake. 

This adds support for sdists releases of this package, git install directly of this repo from pip (submodules may make that hard), and generally simplifies the build process since now all that is required is doing pip install . in the directory or "python -m build" to build the wheel and sdist with no other build steps. This also fixes some cross compiling issues.

This also means that cibuildwheel will work out of the box with github to build this for more architectures and you won't have to maintain as much custom github action CI logic. 

A subset of tests run in CI as part package building now. 

CUDA is optional again which fixes Mac builds. The optional triton kernel is also available as a backend but only when the triton package is also installed in the env.  

Updated documentation and install instructions. 

Add attestation support for uploaded wheels.


Ubospica (commented):
Hi @zbowling, thanks for the contribution! Modernizing the building process with scikit-build-core and cibuildwheel is definitely helpful for XGrammar. We have planned this for a long time, but there is no bandwidth to do it. So thank you for addressing this important issue!

I updated your PR to make it more suitable for XGrammar's specific situation. I'm also just starting to use these tools, so if there are any problems, feel free to point them out.

We also need to consider how to make CD pipeline work after this migration (as the previous CD depends on setup.py). I will try to work that out later, and contributions are welcome!

pyproject.toml
```diff
@@ -44,3 +128,19 @@ select = ["ALL"]
...
 max-args = 10
+
+[tool.cibuildwheel]
+build-verbosity = 1
+skip = ["cp36-*", "cp37-*", "pp*", "*musllinux*"]
+
+# Platform-specific configurations
+[tool.cibuildwheel.linux]
+archs = ["x86_64", "aarch64"]
+# environment = { SKBUILD_BUILD_OPTIONS = "--", CXXFLAGS = "-D_GLIBCXX_USE_CXX11_ABI=0", AUDITWHEEL_PLAT = "manylinux2014_x86_64" }
```
Ubospica: Are the configurations in this line useful?

pyproject.toml
```diff
@@ -44,3 +128,19 @@ select = ["ALL"]
...
+
+[tool.cibuildwheel.macos]
+archs = ["x86_64", "arm64"]
+environment = { SKBUILD_BUILD_OPTIONS = "--", MACOSX_DEPLOYMENT_TARGET = "10.14" }
```
Ubospica: Ditto. Do these configs have any purposes?

pyproject.toml
zbowling: Yeah! Some of this cruft :) I'll clean it up. I sent the draft PR once it started working.

zbowling:
@Ubospica This is getting closer. There was a big change I made here as part of getting the test to work again with CI is fixing the kernels to work when triton and cuda are not available to backback to CPU. Now I'm able to get all tests to pass on MacOS. Some other non-functional changes to clean up error messages to meet lint warnings got mixed in though when I refactored the API there.

zbowling:
The build is green on my repo now! https://github.com/zbowling/xgrammar/actions/runs/13384220521 

We now run a subset of all the unit tests to qualify each wheel during the build (specifically the ones that don't require hugging face tokens). 

This is a bigger PR than I realized.  I fixed several issues with running on MacOS as well. There are a lot of changes overall to fix things to pass tests now on each arch.

zbowling:
Oh, one thing here is I upped the minimum supported version to 3.9 instead of 3.8. We can do 3.8 but we have to stop using certain newer typing features in the tests. 

The other cool thing to maybe to investigate is supporting ABI3/py3/PY_LIMITED abi in the wheel and only building for Py39. It doesn't look like we are using anything outside of the stable ABI. This would reduce the shipped number of wheels and speed up CI, making build faster and something that could run quicker on each PR, although it wouldn't have as much of an impact on users since download size is the same.

python/xgrammar/kernels/apply_token_bitmask_inplace_cuda.py
```diff
@@ -13,52 +13,53 @@
...
-        "-std=c++17",
-        "--threads",
-        "4",
-        "-use_fast_math",
-    ]
-    torch.utils.cpp_extension.load(
+def _load_torch_ops() -> None:
+    with open(__file__.replace(".py", ".cu")) as f:
+        source = f.read()
+    torch.utils.cpp_extension.load_inline(
```
zbowling, line 42: @Ubospica I changed to load_inline because it's safer

python/xgrammar/kernels/__init__.py
```diff
@@ -1,5 +1,26 @@
...
 
+import torch
+
 from .apply_token_bitmask_inplace_cpu import apply_token_bitmask_inplace_cpu
-from .apply_token_bitmask_inplace_cuda import apply_token_bitmask_inplace_cuda
-from .apply_token_bitmask_inplace_triton import apply_token_bitmask_inplace_triton
+
+apply_token_bitmask_inplace = {
+    "cpu": apply_token_bitmask_inplace_cpu,
+}
```
zbowling: This moves to a dictionary of the different kernel impls of this function. If cuda isn't available it's not added to then it's not added to the dictionary. These games for triton where if the optional dependency "triton" isn't installed, it's not added to the dictionary as a possible impl. This fixes Mac builds where cuda doesn't exist and triton can't be installed.

python/xgrammar/kernels/apply_token_bitmask_inplace_cpu.py
```diff
@@ -1,35 +1,60 @@
...
+    @staticmethod
+    def dims_error(tensor_name: str, dims: int) -> str:
+        return f"{tensor_name} should be 1D or 2D, but got {dims}D"
+
+
+LOGITS_CPU_ERROR = "logits must be on CPU"
+BITMASK_CPU_ERROR = "bitmask must be on CPU"
+LOGITS_TYPE_ERROR = "logits must be of type float32"
+BITMASK_TYPE_ERROR = "bitmask must be of type int32"
+MAX_DIMS = 2
```
zbowling: Some of this was just clean up from my linter to get rid of ruff warnings around const strings in errors and having specific error types.

python/xgrammar/matcher.py
```diff
@@ -121,20 +126,49 @@ def apply_token_bitmask_inplace(
...
+        raise DeviceValidationError(
+            DeviceValidationError.device_mismatch(logits.device, bitmask.device)
         )
 
     if logits.device.type == "cuda":
-        if os.environ.get("XGRAMMAR_TOKEN_BITMASK_TRITON") == "1":
-            apply_token_bitmask_inplace_triton(logits, bitmask, indices)
+        if (
+            os.environ.get("XGRAMMAR_TOKEN_BITMASK_TRITON") == "1"
+            and "triton" in apply_token_bitmask_inplace_impl
```
zbowling: this logic will also fall back to cuda if triton isn't installed even if you explicitly set XGRAMMAR_TOKEN_BITMASK_TRITON. Alternative we could throw an error.

scripts/sync_package.py
zbowling, line 1: not needed anymore. The __version__ in version.py is the one truth for version numbers since we have a dynamic=[version] and regex backend to read this field.

tests/python/test_grammar_compiler.py
```diff
@@ -13,6 +13,7 @@
 from xgrammar.testing import _get_allow_empty_rule_ids
 
 
+@pytest.mark.hf_token_required
```
zbowling, line 16: This is a new testing marker I added to flag tests that require a huggingface token with access to the gated models needed in these tests. We skip all these tests at package qualification but I think we still get pretty good coverage. We could run these in CI by passing an HF_TOKEN in an env but then the github runners don't have GPUs so we can't test CUDA paths.

tests/python/test_grammar_matcher.py
```diff
@@ -56,15 +82,46 @@ def test_refuse(input_refused: str):
...
+            4952,
+            128066,
+            128111,
+            4952,
+            128066,
+            128111,
+            4952,
+            127873,
+            128255,
             # fmt: on
```
zbowling: ruff seems to ignore this comment for some reason.

tests/python/test_grammar_matcher_ebnf.py
```diff
@@ -352,7 +520,12 @@ def test_fill_next_token_bitmask(
     print(f"Time to init GrammarMatcher: {(time_end - time_start) / 1e3} us")
 
     token_bitmask = xgr.allocate_token_bitmask(1, tokenizer_info.vocab_size)
-    logits_gpu = torch.zeros(tokenizer_info.vocab_size, dtype=torch.float32, device="cuda")
+    device = "cuda" if torch.cuda.is_available() else "cpu"
```
zbowling, line 350: these tests will fallback to a CPU path if CUDA isn't available now which is nice for MacOS

docs/start/install.rst
```diff
@@ -83,21 +87,28 @@ This step is useful when you want to make modification or obtain a specific vers
 
     # 1. clone from GitHub
     git clone --recursive https://github.com/mlc-ai/xgrammar.git && cd xgrammar
-    # 2. Install pre-commit hooks
+    # 2. Install pre-commit hooks (optional, recommended for contributing to XGrammar)
     pre-commit install
     # 3. build XGrammar core and Python bindings
```
mgorny:
```suggestion
    # 3. build and install XGrammar core and Python bindings
```

docs/start/install.rst
```diff
@@ -73,8 +75,10 @@ This step is useful when you want to make modification or obtain a specific vers
         python=3.11
```
mgorny:
```suggestion
        python=3.11 \
        ninja
```

I don't understand why the original instructions combined `conda` and `pip`, but since you're removing the latter part, the dependencies should probably be installed via `conda` instead (in fact, I dare say they should have been like this since the beginning).

Technically, you could also include `pybind11` <del>and `pytorch`</del> here, and use `--no-build-isolation` in pip<del>, then you'd be able to build against GPU-enabled PyTorch from `conda-forge`</del>.

pyproject.toml
```diff
@@ -1,19 +1,120 @@
...
+provider = "scikit_build_core.metadata.regex"
+input = "python/xgrammar/version.py"
+
+[build-system]
+requires = [
+    "scikit-build-core>=0.10.0",
+    "pybind11>=2.11.0",
+    "wheel",
+    "cmake>=3.18",
+    "ninja",
```
mgorny:
```suggestion
```

scikit-build-core adds these automatically if they are necessary.

pyproject.toml
```diff
@@ -1,19 +1,120 @@
...
+
+[tool.scikit-build.metadata.version]
+provider = "scikit_build_core.metadata.regex"
+input = "python/xgrammar/version.py"
+
+[build-system]
+requires = [
+    "scikit-build-core>=0.10.0",
+    "pybind11>=2.11.0",
+    "wheel",
```
mgorny:
```suggestion
```

I don't think `wheel` is used anywhere. Or at least a quick grep doesn't reveal any imports.

docs/start/install.rst
zbowling: Yeah, it's a somewhat legacy common pattern of folks just using conda just to get a consistent and often newer python toolchain than your distro might provide and have an implicit venv and then get nothing else from conda. 

(Bit of an aside: I'm a conda-forge package maintainer, and ironically the upstream xgrammar feedstock maintainer too, and I prefer to stick to all conda based deps in conda instead of mixing but some folks don't when it comes to python packages and purely just use pypi packages with fine. That is *fine*, as long as you are not mixing and matching conda based python packages and pypi based python packages. Mixing and matching is where you get into complex problems and subtle issues because conda-forge versions are sometimes a bit different in how they are built.)

For this instruction I updated to also show you how you can just create a pure python venv so you don't have to involve conda (assuming you have a decent python installed from somewhere else).

mgorny: Sure, but since Ninja is not a Python package, I think it makes more sense to install it via Conda (and in pair with CMake), rather than have scikit-build pull another one for isolated build venv.

maresb:
Cross-reference to https://github.com/conda-forge/xgrammar-feedstock/pull/5: Consider removing `pybind11` as a runtime dependency.

docs/start/install.rst
zbowling: Ah I see what you are saying.

Ubospica:
Thank a lot @zbowling @maresb! I was too busy the past few days to review it, but I’ll review it tomorrow and get it merged as soon as possible.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,97 @@
...
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - { os: ubuntu-latest,    arch: x86_64,     build: 'cp*-manylinux*' }
+        - { os: ubuntu-24.04-arm, arch: aarch64,    build: 'cp*-manylinux*' }
+        - { os: windows-latest,   arch: AMD64,      build: 'cp*'          }
+        - { os: macos-14,         arch: arm64,      build: 'cp*'          }
```
Ubospica: Could you add MacOS x64 `macos-13` for backward compatibility?

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,97 @@
+name: Build and upload to PyPI
```
Ubospica: This file can be better named build_and_release_pypi.yaml

docs/start/install.rst
```diff
@@ -83,21 +88,28 @@ This step is useful when you want to make modification or obtain a specific vers
...
-    echo "export PATH=\"$(pwd):\$PATH\"" >> ~/.bashrc
 
 **Step 3. Validate installation.** You may validate if XGrammar is compiled successfully in command line.
 You should see the path you used to build from source with:
 
 .. code:: bash
 
    python -c "import xgrammar; print(xgrammar)"
+
+**Step 4. (Optional) Run Python Tests.** You will need a HuggingFace token to run the tests that has gated models.
```
Ubospica:
```suggestion
**Step 4. (Optional) Run Python Tests.** You will need a HuggingFace token and access to gated models to run the tests that have gated models.
```

pyproject.toml
```diff
@@ -1,19 +1,117 @@
...
+    "Intended Audience :: Education",
+    "Intended Audience :: Science/Research",
+]
+keywords = ["machine learning", "inference"]
+requires-python = ">=3.9, <4"
+dependencies = [
+    "pydantic",
+    "sentencepiece",
+    "tiktoken",
+    "torch>=2.5.0",
```
Ubospica: Support old versions of pytorch
```suggestion
    "torch>=1.10.0",
```

pyproject.toml
```diff
@@ -1,19 +1,117 @@
...
-src_paths = ["python"]
+[project.urls]
+Homepage = "https://xgrammar.mlc.ai/"
+GitHub = "https://github.com/mlc-ai/xgrammar"
 
-[tool.mypy]
-strict = true
+[project.optional-dependencies]
+triton = [
+    "triton",
```
Ubospica: Just add triton as a normal dependency (I think it should be fine?)
```
"triton; platform_system == 'Linux' and platform_machine == 'x86_64'"
```

pyproject.toml
```diff
@@ -1,19 +1,117 @@
...
+minimum-version = "build-system.requires"
+
+# Build configuration
+build-dir = "build"
+build.verbose = true
+
+# CMake configuration
+cmake.version = "CMakeLists.txt"
+cmake.args = []
+cmake.build-type = "RelWithDebInfo" # Release is Default
```
Ubospica: Setting RelWithDebugInfo as default is fine.
```suggestion
cmake.build-type = "RelWithDebInfo"
```

pyproject.toml
```diff
@@ -44,3 +142,30 @@ select = ["ALL"]
...
+exclude = "3rdparty/*"
+line-length = 100
+
+[tool.isort]
+profile = "black"
+src_paths = ["python"]
+
+[tool.cibuildwheel]
+build-verbosity = 1
+skip = ["cp36-*", "cp37-*", "cp38-*", "pp*", "*musllinux*", "cp313-manylinux_aarch64", "cp313-macosx_x86_64"] # pypy doesn't play nice with pybind11
```
Ubospica: Why skip cp313-manylinux_aarch64 and cp313-macosx_x86_64?

Ubospica (commented):
I have finished reviewing the build system section. I will check the linting part tomorrow. Overall, the PR looks great. Thanks!

Just a few questions and points to mention:
1. Can this workflow integrate well with the documentation workflow at [documentation.yaml](https://github.com/mlc-ai/xgrammar/blob/main/.github/workflows/documentation.yaml)?
2. Also, thank you for maintaining the wheel on conda-forge. Is there any part of the conda-forge repository that can be moved into the xgrammar repository? If so, it would be ideal if this could help reduce your maintenance workload.
3. Please use the pre-commit hook to format the code. I noticed some extra blank lines, missing newlines at the end of files, and overly long lines.

pyproject.toml
zbowling: python 3.13 on aarch64 and macosx x86 don't have torch versions for torch 2.5.0+. macos x64 is likely never going to never see any new torch builds for any python version after 3.13. The missing Linux aarch64 build for python 3.13 is mostly because python 3.13 is still just in early support for torch. I'll add a comment for the future.

pyproject.toml
zbowling: Sure, but it looks like the code is favoring the direct Cuda version and not the trition version now since you have to set an env to even use triton.

.github/workflows/deploy.yaml
zbowling: sure! Yeah since >=2.3 torch dropped support for Mac x64 I wasn't sure how important it but since we want to support back to torch 1.x that should be easy. There isn't much regression on the older builds but by adding it back, the solver should install torch 2.2.2 on Mac x64 and the rest should get torch 2.6 (except for Python 3.13 on aarch64 where there is no official torch binary yet)

zbowling:
This should look a bit better. Note I swapped out ruff for black in the pre-commit and fixed up some rules (I was using ruff locally in place of black and it's why it blew up some line wraps but I fixed a few things so now it won't do that which reduces a lot of the changes in this PR.

I also noticed the aarch64 and x86_64 linux builds are lot bigger. Specifically because the cuda kernel that gets compiled now I suspect.

image

.pre-commit-config.yaml
```diff
@@ -39,10 +39,15 @@ repos:
       - id: remove-crlf
 
   # Formatters
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 24.1.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
```
Ubospica: Linting is certainly important, but our experience suggests that it should not be enforced in pre-commit; instead, it should be handled in CI. Running linting in pre-commit can prevent committing temporary changes, as it requires fixing all linting errors beforehand.

So we can remove this for now and add it info an workflow later.

pyproject.toml
```diff
@@ -1,19 +1,107 @@
...
+
+  # Tests
+  "/tests/**/*",
+]
+
+sdist.exclude = ["**/.git", "**/.github", "**/__pycache__", "**/*.pyc", "build", "dist"]
+
+# Editable install settings
+# Editables are fairly buggy
+# editable.rebuild = true
```
Ubospica, line 92: Is there any exact issues with editable installation?

python/xgrammar/matcher.py
Ubospica: We can throw an error for this. I have done that in the new commit.

Ubospica (commented):
Hi @zbowling, I just finished the review and I think it is close to being complete now!

python/xgrammar/kernels/apply_token_bitmask_inplace_cpu.py
Ubospica: We can skip the TRY rules for Ruff. There are two reasons:
1. Separating error messages from where they occur could make it harder to maintain
2. We also could throw errors from C++, but the customized exception in Python cannot cover them

pyproject.toml
zbowling: It injects a proxy that tries to rebuild and compile on demand when you import and invoke it. For me, it's not working at all. Something is failing to compile at import time. [Editable native deps are still experimental, so I was going to debug that later.](https://scikit-build-core.readthedocs.io/en/latest/configuration/index.html#editable-installs). Non-editable builds compile more traditionally and more directly, and they work fine. 

I may still poke at fixing it, but I already bit off a ton more here and figured that could be a follow-up PR once I debugged what is broken there.

Ubospica: It's interesting because the editable installation with
```bash
pip install -e --no-build-isolation .
```
works well for me. The rebuilding can be successfully triggered when the C++ codebase is changed.

But of course we can comment this first since there could still be problems. If you have any further findings, I would appreciate it for bringing it up.

zbowling: Oh nice! Yeah I didn't have time to debug the first pass what was failing.

Ubospica:
> For me in that conda-forge recipe, I can get around it by changing include paths to include to look at $PREFIX/include/dlpack where dlpack gets install in the dev env instead of 3rdparty/dlpack in the recipe.
> 
> One change we could do is a find_package() module search for dlpack in cmake before falling back to searching in 3rdparty/, and then I wouldn't have to inject any header search paths.

Got it! We can do that in the next PR.

> Picojson was another but this one gets an exception in conda-forge since picojson upstream has been unmaintained for several years and many different ML projects have their own hard forks of it. Googletest isn't used in the conda-forge build so I largely ignore it.

Indeed picojson has been modified a lot by us. I think it can now be treated as part of the codebase rather than an external dependency, so there’s no need to rely on the upstream Conda package.

I think this PR is ready and I will merge it. Thanks a lot for your help @zbowling !

tests/python/test_grammar_matcher.py
```diff
@@ -106,13 +96,13 @@ def test_fill_next_token_bitmask(
         assert rejected_sizes[-1] == expected_rejected_sizes[-1]
 
 
-def test_token_operations() -> None:
+def test_token_operations():
     """Test accepting token and finding the next token mask."""
-    # fmt: off
     vocab = [
+        # fmt: off
```
zbowling, line 102: I changed these to be outside the statement because Ruff ignores them inside the statement. https://docs.astral.sh/ruff/formatter/#format-suppression

pyproject.toml
zbowling: Oh interesting. Editable builds work on my Linux machine but not my Mac. Debugging this stack trace

tests/python/test_grammar_matcher.py
Ubospica: Thanks for bringing that up. But since we use black as the formatter instead of ruff, and black supports `fmt: off` by lines, this should be fine as well.

pyproject.toml
Ubospica: I see, yeah it should work on Linux. We may need to see how Mac supports it.

Ubospica: @zbowling Just wanna ask if this editable installation works now. I feel it is helpful in the real development, so if it has no compatibility issues, we can enable it.

reneleonhardt:
```
wheels are missing for Linux aarch64 for pytorch so temporarily skip cp313-manylinux_aarch64
```
pytorch builds manylinux aarch64 wheels for Python 3.10 - 3.14, including free-threaded 3.13t - 3.14t.
https://pypi.org/project/torch/2.9.0/#files

And because dependabot isn't used here, cibuildwheel@v2.22.0 (1 year old) also didn't build any Python 3.14 wheels 3 days ago.
So, no wheels for Python 3.8 and 3.14 at all, and Python 3.13 Linux aarch64 despite pyproject.toml claiming `>=3.8, <4`:
https://pypi.org/project/xgrammar/0.1.26/#files

PR #197, opened 2025-02-13
Fix potential maybe-uninitialized warning

syuoni:
I encountered "-Werror=maybe-uninitialized" when compiling the latest version of xgrammar in TensorRT-LLM. The warning can be resolved by explicitly checking if `root_tag_dispatch_fsm` is initialized.

In xgrammar's CMake file, setting `-flto=auto` can also prevent this warning. 

Do you mind merging this PR? @Ubospica

### DIFF
--- a/cpp/grammar_matcher_base.cc
+++ b/cpp/grammar_matcher_base.cc
@@ -14,6 +14,7 @@
 #include "grammar_data_structure.h"
 #include "persistent_stack.h"
 #include "support/encoding.h"
+#include "support/utils.h"
 
 namespace xgrammar {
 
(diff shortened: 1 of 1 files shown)


Ubospica:
Hi @syuoni , thanks for pointing that out! Could you provide your GCC or Clang version and compilation flags? I didn't encounter this compilation error on my side, so understanding your environment may be able to help solve this problem.

This PR also looks good to me. We can merge it soon.

syuoni:
Hi @Ubospica , removing the `-flto=auto` flag from XGrammar's CMakeLists.txt can reproduce the warning.

Ubospica:
Thanks for the explanation @syuoni! That makes sense. May I also ask if you have any reason to turn off auto lto during deployment? If so I can remove this or pay attention to this to avoid future compilation problems.

syuoni:
Hi, actually I don't think it's a problem of XGrammar itself.

The project I'm working on has very complicated compilation flags to support different platforms, cpu archs, C++ ABIs, so I directly compile XGrammar's source files with these compilation flags, instead of using XGrammar's compilation flags. 

I will talk to team members to see if we can enable `-flto=auto`. While, I would still greatly appreciate if you can merge this change to XGrammar.

Ubospica:
@syuoni I see, then I think we can keep this compilation option and merge this pr. Thanks for your contribution!

PR #217, opened 2025-02-25
Update README to announce that Modular MAX uses XGrammar

zbowling:
Modular MAX 25.1 shipped this month, which uses XGrammar for structured output. 

https://docs.modular.com/max/serve/structured-output


Ubospica:
Sounds great, thanks!

Issue #214, opened 2025-02-24
Add CI for XGrammar


Ubospica:
XGrammar needs CI to ensure stability and help review community contributions. Guidelines for CI:
1. Currently CI only needs to run on CPU (tests requiring GPU can be skipped, specifically the GPU kernel tests)
2. Run linter (pre-commit hooks, ruff)
3. Test on multiple platforms (Linux, MacOS, Windows) x (x64, ARM64)
4. Test C++ and Python code (ctest and pytest)
5. CI should be enforced to every commit on main branch, every PR, and can be manually triggered
6. The model folder `~/.cache/huggingface` should be shared between different runs
Contribution is welcome!

Ubospica:
See #241 #218

PR #218, opened 2025-02-25
[Feature] Build and run C++ Python tests

DarkSharpness:
Add ci (#214) to build xgrammar from source code and run all C++ tests and Python tests (except for those HF_TOKEN needed tests).


DarkSharpness:
Update: Add support for multi-arch

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Build xgrammar from source
+        run: |
+          echo "set(XGRAMMAR_BUILD_CXX_TESTS ON)" >> cmake/config.cmake
+          python -m pip install --upgrade pip
+          pip install -e .
+
+      - name: Run C++ tests
+        run: |
+          cd build && ctest && cd ..
```
Ahajha: Should be able to do a `ctest --test-dir build`

DarkSharpness: Thanks!

zbowling (approved):

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,55 @@
...
+          pip install pytest
+          pytest
+
+      - name: Run Python tests without HF_TOKEN
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        if: env.HF_TOKEN == ''
+        run: |
+          pip install pytest
+          pytest -m "not hf_token_required"
```
Ubospica, line 55: Is this part necessary? I think `Run Python tests` should have covered all tests?

Ubospica (commented):
Thanks for the contribution! It looks good to me.

.github/workflows/unit_test.yaml
DarkSharpness: Personally, I think those without HF tokens should still run the basic tests without HF token requirements.

Ahajha: If the tests are just for this repo, wouldn't we always have HF_TOKEN?

Ubospica: For this repo, I think so. But this dispatching would be helpful for forked repos because they may not have this HF_TOKEN defined.

PR #219, opened 2025-02-25
[Style] Remove unused headers

DarkSharpness:
Currently, there are many headers in C++ files that are not used directly. We remove all of them (thanks to clangd).

Moreover, it seems that an `inline` is missing in `cpp/support/thread_pool.h`, which may lead to a potential violation of ODR in the future.

Maybe this PR should be merged only after a C++ test ci is ready to ensure correctness.

### DIFF
--- a/cpp/grammar.cc
+++ b/cpp/grammar.cc
@@ -5,7 +5,6 @@
 
 #include <xgrammar/grammar.h>
 
-#include "grammar_data_structure.h"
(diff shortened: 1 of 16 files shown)


Ubospica:
Is IWYU or clangd robust enough to fit into CI/pre-commit-hook pipeline? Or we can manually run it once in a while.

Ubospica (approved):
LGTM!

PR #223, opened 2025-02-26
Fallback to triton if we fail to compile for CUDA

zbowling:
The CUDA driver might be available but the full CUDA toolkit, ninja, and gcc/clang might not be installed so we won't have enough tools to compile the kernel, and in this case use triton as a fallback. 

We could consider doing an AOT of the kernel and not JITing the kernel extension, however this really blows the packaging complexity out of the water since we would have to target multiple PyTorch extension ABIs.

### DIFF
--- a/python/xgrammar/kernels/__init__.py
+++ b/python/xgrammar/kernels/__init__.py
@@ -8,10 +8,17 @@ apply_token_bitmask_inplace_kernels = {"cpu": apply_token_bitmask_inplace_cpu}
 
 __all__ = ["apply_token_bitmask_inplace_kernels"]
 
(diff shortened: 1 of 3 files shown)


zbowling:
hey @Ubospica do mind taking a look at these changes? We are stuck on 0.1.11 because of this where we don't have a full CUDA toolkit for runtime JIT but do have a CUDA driver

Ubospica (approved):
@zbowling Thanks for the contribution! Sorry for the late reply, as I was quite busy last week. It looks great to me.

zbowling:
awesome!

PR #230, opened 2025-03-07
Convert from pybind11 to nanobind

Ahajha:
See nanobind: https://github.com/wjakob/nanobind tl;dr Smaller, faster compiled extensions. Nanobind is written by the original author of pybind11, with the hindsight and lessons from pybind11.

Locally (x64 Linux), the compiled extension went from 14.10MB to 12.01MB.

I have some very crude benchmarks in the comments - this is performing about on par with the previous code.

I've run all the tests locally (x64 Linux), everything passes. Ideally #218 is merged so we can test those automatically on other platforms, but the only major differences I would expect are binary sizes.

I make some CMake changes so that the compiled extensions now just depend on the core project, rather than rebuilding all the sources again for the extension.


Ahajha:
Something to note that I'm not sure about - I _think_ pypy works with nanobind, so I might try re-adding it.

Ubospica:
@Ahajha Thanks for the contribution! Nanobind looks great and fast. 

Regarding the performance, could you check the output of 
```bash
pytest tests/python/test_grammar_matcher_ebnf.py::test_fill_next_token_bitmask
```
before and after this PR? As long as there is no major performance degradation, we can migrate nanobind.

Besides, could you run `pre-commit install && pre-commit run` to format the modified files? I noticed several files don’t seem to be formatted correctly, and it should be very easy to format them with pre-commit hooks.

Ahajha:
Sure, I'll run that tomorrow. For now I tried running the commands for pre-commit and got the following error, any idea what's going on? I'm not familiar with it:
```
(.venv) alex@alex-fw:~/Documents/xgrammar$ pre-commit install && pre-commit run
pre-commit installed at .git/hooks/pre-commit
An error has occurred: InvalidManifestError: 
==> File /home/alex/.cache/pre-commit/repogfgd3t0n/.pre-commit-hooks.yaml
==> At Hook(id='check-added-large-files')
==> At key: stages
==> At index 0
=====> Expected one of commit, commit-msg, manual, merge-commit, post-checkout, post-commit, post-merge, post-rewrite, prepare-commit-msg, push but got: 'pre-commit'
Check the log at /home/alex/.cache/pre-commit/pre-commit.log
```

Ahajha:
Got some of the HF tests running, ran into a snag due to a few places allowing interchange between bytes and strings. Nanobind specifically doesn't like this (see https://github.com/wjakob/nanobind/discussions/137), so I might need to make some slightly deeper API changes to get a clean solution.

Ahajha:
The issue is with `TokenizerInfo.__init__` (the `encoded_vocab` parameter), which expects either a list of strings or a list of bytes, but then in C++ these both get cast to a vector of strings. Nanobind doesn't make this conversion, but I would guess that we probably want to actually model these as one type or the other rather than either?

Ahajha:
I came up with a solution that I think is reasonable - basically we defer casting until we get into the relevant functions (there was one other that needed fixing). So instead of directly taking in a vector of strings, we just take the raw python list (with some type annotations, these don't do any runtime checks) and then manually cast from either a string or bytes. This solution ended up being fairly simple without requiring any API changes.

Ahajha:
Hmm, I'm seeing a slowdown in the test you asked. I'm running
```
/usr/bin/time pytest tests/python/test_grammar_matcher_ebnf.py::test_fill_next_token_bitmask >/dev/null
```
For main after 3 runs I get:
```
5.88user 0.32system 0:04.10elapsed 151%CPU (0avgtext+0avgdata 722716maxresident)k
5.83user 0.35system 0:04.13elapsed 149%CPU (0avgtext+0avgdata 722200maxresident)k
5.76user 0.31system 0:04.10elapsed 148%CPU (0avgtext+0avgdata 723288maxresident)k
0inputs+232outputs (0major+143446minor)pagefaults 0swaps
```
For my branch I get:
```
6.59user 0.36system 0:09.72elapsed 71%CPU (0avgtext+0avgdata 722100maxresident)k
6.11user 0.33system 0:04.19elapsed 153%CPU (0avgtext+0avgdata 722176maxresident)k
6.31user 0.32system 0:04.36elapsed 152%CPU (0avgtext+0avgdata 722468maxresident)k
```
Something seems up, so I will investigate.

Ahajha:
After some investigating and fixes, I have some better numbers, similar testing method as before.
Main:
```
5.44user 0.25system 0:03.61elapsed 157%CPU (0avgtext+0avgdata 722316maxresident)k
5.64user 0.29system 0:03.79elapsed 156%CPU (0avgtext+0avgdata 723012maxresident)k
5.34user 0.27system 0:03.47elapsed 161%CPU (0avgtext+0avgdata 722360maxresident)k
5.39user 0.28system 0:03.51elapsed 161%CPU (0avgtext+0avgdata 722292maxresident)k
5.15user 0.27system 0:03.40elapsed 159%CPU (0avgtext+0avgdata 723596maxresident)k
5.30user 0.24system 0:03.40elapsed 162%CPU (0avgtext+0avgdata 723056maxresident)k
```
Nanobind branch:
```
5.37user 0.32system 0:04.05elapsed 140%CPU (0avgtext+0avgdata 721436maxresident)k
5.41user 0.26system 0:03.58elapsed 158%CPU (0avgtext+0avgdata 723240maxresident)k
5.35user 0.28system 0:03.53elapsed 159%CPU (0avgtext+0avgdata 722456maxresident)k
5.40user 0.27system 0:03.53elapsed 160%CPU (0avgtext+0avgdata 722348maxresident)k
5.44user 0.28system 0:03.66elapsed 156%CPU (0avgtext+0avgdata 723440maxresident)k
```
This is definitely eyeball statistics, but this seems to be on par with what it is now. I expected a speedup, perhaps there is more to investigate, but as you said as long as it isn't an obvious regression it should be okay. I also have some future PR ideas that involve nanobind that should lead to further speedups (some of them could be done with pybind11, but will be easier with nanobind).

Ahajha:
The issue was actually rather insightful, I'll share some insight for you (or anyone else reading this PR).

The first issue was that pybind11 enables LTO by default, nanobind doesn't - the former basically requires it to get reasonable performance, but nanobind is designed and laid out in a better way so that you can still get good performance without it, though they provide the option should we want it. I know this project has `-flto=auto` on by default, I'm not sure why this sped things up but it did. Perhaps it's a different level of LTO, it's whatever CMake's `CMAKE_INTERPROCEDURAL_OPTIMIZATION` does. 

The second issue was more interesting to me. Nanobind has a section on their CMake API (https://nanobind.readthedocs.io/en/latest/api_cmake.html#command:nanobind_add_module), and one thing they mention there is binary size. Nanobind defaults to optimizing for _size_, because in the author's experience with bindings, optimizing for speed tends to not make a difference. However they do make a note about mixing optimization levels, specifically, compiling the core project with whatever default optimization level like -O2, and just binding code with -Osize. So I've slightly rearranged some code to make anything that is just interop (with no 'real' logic) into nanobind.cc, no other files depend on nanobind.

Ahajha:
@Ubospica This should be ready for another look now!

zbowling (approved):

Ubospica:
Hi @Ahajha, thank you so much for your contribution! I just find time to review it. I will finish the review soon!

Ubospica (approved):
This PR looks great to me! Just a few comments and questions:

1. >  I expected a speedup, perhaps there is more to investigate, but as you said as long as it isn't an obvious regression it should be okay. I also have some future PR ideas that involve nanobind that should lead to further speedups (some of them could be done with pybind11, but will be easier with nanobind).

I tested the code and also feel the performance improvement here is relatively small (if I did not make it wrong). One reason may be that we make the binding very lightweight, so there is not much interaction between C++ and Python. But it is great in terms of reducing the binary size. Also, any further contribution is welcome!

2. > I know this project has -flto=auto on by default, I'm not sure why this sped things up but it did

Did you observe a speedup when enabling lto? We enabled it by default and did not consider it thoroughly. It would be great if it is indeed helpful.

The entire code is very clear, and I feel we can merge it soon!

Ahajha:
I'll look into the Windows issue shortly, and respond to your questions.

Ahajha:
@zbowling Do you have any idea what's going on with this? https://github.com/mlc-ai/xgrammar/actions/runs/13869260346/job/38843092767?pr=230#step:4:25074

pybind11 has a rather complicated Python finding method, nanobind just uses FindPython, which seems to mostly work but it isn't finding some of the other versions on Windows for some reason.

Ahajha:
RE: LTO, without nanobind's LTO option there was a bit of a slowdown, with it I saw it come back to parity (which I think you're correct that there just isn't much overhead). There might be some investigation to be done regarding the LTO that nanobind is doing vs what we're doing.

Ahajha:
After some digging, I asked a question in the pypa discussion board [here](https://github.com/pypa/cibuildwheel/discussions/2318), hopefully we'll get some insight. Not sure if it's a bug on their side or if we're doing something wrong.

cpp/pybind/CMakeLists.txt
```diff
@@ -1,31 +0,0 @@
-set(PYBIND11_NEWPYTHON ON)
```
henryiii: You were using FindPython before, too, that's that this setting does.

Ahajha: Hmm, does pybind11 set any extra flags to help Find Python, like extra search paths or hints?

henryiii: Yes, but the ones I know of should only affect PyPy, and only for stand-alone builds (scikit-build-core makes sure FindPython works on PyPy). Do you have logs for a pybind11 build? What might be happening is that the library was never found, but nanobind needs it because it's generating a list of symbols to ignore, while pybind11 just ignores all unresolved symbols.

Ahajha: Log here: https://github.com/mlc-ai/xgrammar/actions/runs/13894029090/job/38870730963- the raw log is [here](https://productionresultssa0.blob.core.windows.net/actions-results/86186adc-905f-4a32-9927-0ce36a5ee561/workflow-job-run-f163f28b-5b5d-512e-953a-98b8f90fe4e8/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-03-17T14%3A58%3A48Z&sig=8uPfYZmCWMDJ9u%2F%2Bk4NYecDn%2BYOZ6xGnpC3kteyg79c%3D&ske=2025-03-18T02%3A07%3A33Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-03-17T14%3A07%3A33Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-01-05&sp=r&spr=https&sr=b&st=2025-03-17T14%3A48%3A43Z&sv=2025-01-05), you can search "CPython 3.10 Win" to find the 3.10 job starting. This was the most recent pybind11 build, 7 hours ago.

henryiii: Strange, it is using the old search, though it looked like it was told not to in the CMakeLists:

```
2025-03-17T07:43:40.6190293Z 2025-03-17 07:43:40,605 - scikit_build_core - WARNING - Can't find a Python library, got libdir=None, ldlibrary=None, multiarch=None, masd=None
...
packages/pybind11/share/cmake/pybind11/FindPythonLibsNew.cmake:101 (message):
2025-03-17T07:43:42.3823327Z   Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
2025-03-17T07:43:42.3823878Z   are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
2025-03-17T07:43:42.3824401Z   the cmake_policy command to set the policy and suppress this warning, or
2025-03-17T07:43:42.3824930Z   preferably upgrade to using FindPython, either by calling it explicitly
2025-03-17T07:43:42.3825443Z   before pybind11, or by setting PYBIND11_FINDPYTHON ON before pybind11.
2025-03-17T07:43:42.3825993Z Call Stack (most recent call first):
2025-03-17T07:43:42.3826758Z   C:/Users/runneradmin/AppData/Local/Temp/build-env-4wr144g5/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Tools.cmake:50 (find_package)
2025-03-17T07:43:42.3827961Z   C:/Users/runneradmin/AppData/Local/Temp/build-env-4wr144g5/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Common.cmake:228 (include)
2025-03-17T07:43:42.3829219Z   C:/Users/runneradmin/AppData/Local/Temp/build-env-4wr144g5/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Config.cmake:250 (include)
2025-03-17T07:43:42.3829963Z   cpp/pybind/CMakeLists.txt:2 (find_package)
2025-03-17T07:43:42.3830346Z This warning is for project developers.  Use -Wno-dev to suppress it.
2025-03-17T07:43:42.3830646Z 
2025-03-17T07:43:42.4102626Z -- Found PythonInterp: C:\Users\runneradmin\AppData\Local\Temp\build-env-4wr144g5\Scripts\python.exe (found suitable version "3.10.11", minimum required is "3.7")
2025-03-17T07:43:42.4383412Z -- Found PythonLibs: C:/Users/runneradmin/AppData/Local/pypa/cibuildwheel/Cache/nuget-cpython/python.3.10.11/tools/libs/python310.lib
```

I notice that the libraries are found outside of the venv, in the parent environment. Interesting. I'd first like to see if we can get the new search to run on pybind11.

Ahajha: Looking at your logs from your test (https://github.com/henryiii/xgrammar/pull/1) - interesting that it seems to work without issue.

Ahajha: I figured out the issue, and it seems kind of strange that this would affect it, but I needed to add the `Interpreter` component. Once I added that things went smoothly. (test [here](https://github.com/Ahajha/xgrammar/actions/runs/13914659118).

Thanks for looking, that gave a good starting point of where else to check!

Ahajha:
@Ubospica Windows issue is fixed!

Ahajha:
Ah looks like I need to run precommit, will do that shortly

cpp/pybind/CMakeLists.txt
Ubospica: That makes sense. We also included the `Interpreter` prior to migrating to `set(PYBIND11_NEWPYTHON ON)` because the latter seems more convenient.

Thank you both for digging here, and I'm glad this is finally resolved!

henryiii: Yes, you need Interpreter for it to ask the Interpreter where its libs are. Not great for cross-compiling, though, so I'll see if I can get scikit-build-core to report the correct lib location in this case.

Ubospica:
Thanks! We can merge it soon after the CI passes.

pyproject.toml
```diff
@@ -35,7 +35,7 @@ provider = "scikit_build_core.metadata.regex"
 input = "python/xgrammar/version.py"
 
 [build-system]
-requires = ["scikit-build-core>=0.10.0", "pybind11>=2.11.0"]
+requires = ["scikit-build-core>=0.10.0", "nanobind>=2.5.0"]
```
Ubospica: It would also be great if you could check the compatibility with nanobind>=2.0.0. I checked the changelog of nanobind and it seems to work. But it would be best if you could check it out. Compatibility with the lower version of nanobind will help users manage library versions in the environment.

Ahajha: That makes sense. I just grabbed the latest - I'll try some earlier versions and see how things go.

Ahajha: One minor change needed, but otherwise it works with 2.0.0. I had to remove `NB_SUPPRESS_WARNINGS`, which means the builds will be a little noiser due to warnings from building nanobind. Can just add a check for the nanobind version to see whether to add it or not if we wanted, for now I went for the simple option and we can revisit later. That flag is new in 2.5.0 anyways.

Ubospica: Sounds great. I agree that we can revisit it later.

Ubospica:
@Ahajha Thanks for your contribution!

cpp/pybind/CMakeLists.txt
Ubospica: @henryiii Yeah, that is rather important. Glad to hear if you have any updates on this!

PR #242, opened 2025-03-14
[Fix] Add support for kDLCUDAHost in token bitmask device type check

Irfnfnkemed:
- Modify the device type check to allow kDLCUDAHost in addition to kDLCPU.
- This change ensures that token bitmasks can be stored in pinned memory (kDLCUDAHost).


Ubospica (approved):
LGTM. Thanks!

PR #249, opened 2025-03-17
fix(python): use correct name for pybind11 FindPython mode

henryiii:
See #230. Want to see if this makes it through or not.


Ahajha:
This makes sense, I was looking for "PYBIND11_NEWPYTHON" in the source and couldn't find anything. If you want to test the specific CI run that triggers this without waiting for approval, you can modify this line (https://github.com/mlc-ai/xgrammar/blob/main/.github/workflows/build_and_release.yaml#L8) and the CI will run on your fork.

henryiii:
I can do this, avoids pushing a change: https://github.com/henryiii/xgrammar/pull/1

PR #255, opened 2025-03-19
[Test] Add more test environments.

Seven-Streams:
- Add more test environments. Instead of only running with python 3.11, now tests can run with from python3.8 to python3.13. However, macos-13-x86_64, python13 is excluded, since pip can't find any torch distribution in this case. 
- Add caches for pip to accelerate the workflow.


Ubospica (approved):
LGTM. Thanks!

PR #257, opened 2025-03-19
[CI]: add a yaml to manually remove all the caches.

Seven-Streams:
As the title describes, add a yaml to manually remove all the caches.


Ubospica:
It seems this can be achieved through github web UI. So a separate action for that may not be necessary. See also

<https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#viewing-cache-entries>

PR #258, opened 2025-03-20
[Fix] Fix primitives for json schema converter

Ubospica:
This PR solves the "root rule not found" issue for primitives in json schema converter.


Ubospica:
#264

PR #259, opened 2025-03-20
[CI] remove tests in build_and_release.yaml.

Seven-Streams:
Since our wheels are tested in unit_test.yaml, there is no need to test wheels in build_and_release.yaml, which is time-consuming.


Ubospica:
Thanks! We can do it in #262.

PR #261, opened 2025-03-20
Fix CI

Ubospica:


Ubospica:
See #262 which moves the branch into the main repo to get access to secrets.

PR #264, opened 2025-03-21
[Fix] Fix primitives for json schema converter

Ubospica:
This PR solves the "root rule not found" issue for primitives in json schema converter.

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -739,6 +739,11 @@ std::string JSONSchemaConverter::CreateRuleFromSchema(
 ) {
   std::string idx = GetSchemaCacheIndex(schema);
   if (basic_rules_cache_.count(idx)) {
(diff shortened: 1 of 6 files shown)


cpp/json_schema_converter.cc
```diff
@@ -739,6 +739,11 @@ std::string JSONSchemaConverter::CreateRuleFromSchema(
 ) {
   std::string idx = GetSchemaCacheIndex(schema);
   if (basic_rules_cache_.count(idx)) {
+    if (rule_name_hint == kRootRuleName) {
+      // If the rule name is root, we need to define the root rule instead of just using the
+      // cached rule.
+      return ebnf_script_creator_.AddRule(rule_name_hint, basic_rules_cache_[idx]);
+    }
```
wangkuiyi, line 746: Would this change the behavior of the converter and make it tend to generate a new rule? @Ubospica

PR #267, opened 2025-03-23
update ci for hf token

zanderjiang:
Adds hf token so the benchmark can correctly run the Llama-3.1-8B-Instruct tokenizer


Ubospica:
Could you merge with main?

Issue #270, opened 2025-03-25
Should we support MLX and Apple Silicon?


wangkuiyi:
I have a simple example running successfully on my Mac Studio (M2) and other Apple Silicon machines.

It demonstrates that XGrammar works with MLX just as it does with PyTorch, and with mlx-lm in much the same way it works with Hugging Face Transformers. From here on, we could add an Apple Silicon GPU kernel for MLX just like the existing CUDA/Triton kernel for PyTorch.

Would this be something appropriate for a pull request to merge into the project? I ask because it would add mlx-lm as a dependency in the pyproject.toml file and might also require adding Apple Silicon as a target in the CI system.

I’d appreciate your guidance—if you think it’s a good fit, I’d be happy to submit a pull request. And if not, I completely understand.

Thanks so much!

```python
"""
Usage:
    python xgrammar.py --model mlx-community/Qwen2.5-Coder-32B-Instruct-3bit
"""

import argparse
import mlx.core as mx
from mlx_lm.generate import generate as mlx_generate
from mlx_lm.utils import load as mlx_load
from transformers import AutoTokenizer
import xgrammar
import torch


def apply_token_bitmask_inplace_mlx(
    bitmask: torch.Tensor, logits: mx.array
) -> mx.array:
    """This is an easy mimic of the apply_token_bitmask_inplace function.

    Args:
        bitmask (torch.Tensor): Created by calling xgrammar.allocate_token_bitmask.
        logits (mx.array): Passed in from mlx_generate.

    Returns:
        mx.array: The masked logits where invalid tokens have their logits set to -inf
    """
    vocab_size = logits.shape[-1]
    # Create mask as torch tensor for mutability
    logits_mask = torch.zeros(logits.shape, dtype=torch.float32)
    for i in range(bitmask.size(0)):
        mask_int = bitmask[0][i].item()  # batch size is 1
        for bit in range(32):
            token_idx = i * 32 + bit
            if token_idx >= vocab_size:
                break
            is_valid = (mask_int >> bit) & 1
            logits_mask[0][token_idx] = 0.0 if is_valid else float("-inf")
    # Convert to mx.array only so we can add it to logits
    return logits + mx.array(logits_mask.numpy())


class XGrammarLogitsProcessor:
    def __init__(
        self,
        grammar: xgrammar.CompiledGrammar,
        max_rollback_tokens: int = 16,
    ):
        self.matcher = xgrammar.GrammarMatcher(
[...]

PR #273, opened 2025-03-25
Add module xgrammar.contrib.mlxlm

wangkuiyi:
After a chat with @Ubospica , I am filing this PR to fix https://github.com/mlc-ai/xgrammar/issues/270

This module provides a logit processor for mlx_lm, just like hf.py provides one for Hugging Face Transformers.

This module is runnable and works demonstrating that XGrammar runs on Apple Silicon via MLX and mlx-lm, just like it works on CUDA via PyTorch and Hugging Face Transformers.

You can run this module like the following

```shell
python -m mlxlm \
 --model mlx-community/Qwen2.5-Coder-3B-Instruct-4bit \
 --prompt "Generate a very simple example JSON. No text. Only the JSON."
```

It generates a grammar-verified JSON like the following:

```json
{
  "name": "John Doe",
  "age": 30,
  "city": "New York"
}
```

### Optional Dependency

This pull request adds mlx-lm as an optional dependency `apple-silicon`. To run this demo modules, you need to install XGrammar from source code:

```shell
pip install -e .[apple-silicon]
```

### Next Steps

1. Currently, this demo module expand the bitmask generated from the pushdown automaton of XGrammar into a float array before adding it to the vocabulary logits generated by the LLM. We should add a Apple Silicon GPU kernel for MLX, just like the CUDA, Triton and CPU kernels for PyTorch.  In this way, we don't expand the bitmask and saves application time.

2. Currently, XGrammar generates the bitmask as a PyTorch tensor and this demo module converts it into an MLX array. I haven't benchmarked the cost of the conversion. If it is high, we may consider enabling XGrammar generating the bitmask as an MLX array optionally.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ test = [
   # https://github.com/huggingface/transformers/issues/36906
   "transformers<4.50.0; platform_system == 'Darwin'",
 ]
+apple-silicon = ["mlx-lm"]
 
 [tool.scikit-build.metadata.version]
 provider = "scikit_build_core.metadata.regex"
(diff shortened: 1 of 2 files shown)


Ubospica (approved):
LGTM. Thank you so much for your contribution! @wangkuiyi

PR #283, opened 2025-03-28
Add MLX CPU/GPU kernels to support Apple Silicon

wangkuiyi:
This PR follows up https://github.com/mlc-ai/xgrammar/pull/273

PR #285, opened 2025-04-01
Add MLX Metal GPU kernel

wangkuiyi:
## Motivation

The goal of this PR is to enable XGrammar to work with MLX, in addition to its existing support for PyTorch.

## Approach 1: ✅ The Merged

https://github.com/mlc-ai/xgrammar/pull/273

@Ubospica has merged my initial implementation in [PR #273](https://github.com/mlc-ai/xgrammar/pull/273).

This version converts the bitmask generated by the pushdown automaton from a torch.Tensor to a mx.array float mask before applying it to logits: mx.array returned from an LLM via mlx-lm.

While functional, this approach is likely less efficient than the CUDA/Triton version, for two reasons:

1. It expands the bitmask into a float mask.
2. It applies the mask out-of-place, creating a new tensor instead of modifying logits in-place.

## Approach 2: ⛔ The Blocked

https://github.com/mlc-ai/xgrammar/pull/283

As an improvement, I submitted [PR #283](https://github.com/mlc-ai/xgrammar/pull/283), which added custom MLX kernels for ARM CPU and Metal GPU using CMake.

However, building the Metal kernel this way turned out to be very complex—likely due to the intricate relationship between CMakeLists.txt and the pip build backend.

@Ubospica kindly suggested exploring [mlx.core.fast.metal_kernel](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.fast.metal_kernel.html), which supports JIT-compilation of Metal kernels—similar to how torch.utils.cpp_extension.load_inline works for CUDA.

## Approach 3: 🟡 The Imperfect

https://github.com/mlc-ai/xgrammar/pull/285

This PR adopts @Ubospica’s suggestion. It works as expected with the following:

```shell
pip uninstall -y xgrammar && pip install .[test]
pytest -v tests/python/test_token_bitmask_operations.py::test_apply_token_bitmask_inplace
```

Result:

```plaintext
PASSED tests/python/test_token_bitmask_operations.py::test_apply_token_bitmask_inplace[cpu]
PASSED tests/python/test_token_bitmask_operations.py::test_apply_token_bitmask_inplace[metal]
SKIPPED [1] tests/python/test_token_bitmask_operations.py:43: CUDA is not installed
SKIPPED [1] tests/python/test_token_bitmask_operations.py:45: Triton is not installed
```

However, `mlx.core.fast.metal_kernel` allocates a new output tensor instead of performing the operation in-place. This behavior is reasonable—MLX follows a functional programming paradigm similar to JAX, where tensors are immutable. This differs from PyTorch, which allows in-place operations.

## Approach 4: 🚀 The Promising

https://github.com/mlc-ai/xgrammar/pull/287

After discussing with Awni Hannun and Angelos Katharopoulos, they suggested a much better direction:

1. Don’t perform the masking in-place.
2. Expand the bitmask into a float mask (as in Approach 1).
3. Use mx.compile to optimize and accelerate this expand-and-apply process.

Angelos even provided a brilliant and fully compiled implementation:

```python
@mx.compile
def apply_bitmask(bitmask, logits):
    bitmap = mx.array(
        [l[::-1] for l in product(*[[float("-inf"), 0]]*8)],
        dtype=logits.dtype,
    )
    bitmask = bitmask.view(mx.uint8)
    return logits + bitmap[bitmask].flatten(-2)[..., :logits.shape[-1]]
```

My Interpretation of Angelos’ Code:
1. `bitmap` is a (256, 8) array. Each row represents one of the 256 combinations of 8 bits, mapping bits to -inf or 0.
2. `bitmask.view(mx.uint8)` casts the uint32 tensor into uint8 without changing the byte order. (Apple Silicon is little-endian.)
3. As `bitmask` has shape (batch_size, ceil_div(vocab_size, 32)/8), then `bitmap[bitmask]` becomes `(batch_size, ceil_div(vocab_size, 32)/8, 8)`. After `flatten(-2)`, we get a float mask of shape `(batch_size, vocab_size_padded)`.
4. `[..., :logits.shape[-1]]` truncates the padding to match the actual vocabulary size.

## Performance Comparison

Angelos benchmarked approach 4 against my original (non-compiled) Approach 1:

```
In [3]: %timeit mx.eval(apply_token_bitmask_inplace_mlx(bitmask_torch, logits))
255 ms ± 2.47 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %timeit mx.eval(apply_bitmask(bitmask_mlx, logits))
225 µs ± 15.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

**Result: The compiled version is over 1000× faster.**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,7 +48,6 @@ repos:
     rev: 6.0.0
     hooks:
       - id: isort
-        args: [--profile, black]
 
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v19.1.7
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,6 +48,7 @@ repos:
     rev: 6.0.0
     hooks:
       - id: isort
+        args: [--profile, black]
 
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v19.1.7
--- a/python/xgrammar/contrib/mlxlm.py
+++ b/python/xgrammar/contrib/mlxlm.py
@@ -4,6 +4,8 @@ Usage:
 """
 
 import argparse
+import itertools
+import timeit
 
 import mlx.core as mx
 import torch
@@ -27,9 +29,9 @@ def apply_token_bitmask_inplace_mlx(bitmask: torch.Tensor, logits: mx.array) -> 
     Returns:
         mx.array: The masked logits where invalid tokens have their logits set to -inf
     """
-    # If Metal implementation is available, use it
-    if "metal" in xgrammar.kernels.apply_token_bitmask_inplace_kernels:
-        return xgrammar.kernels.apply_token_bitmask_inplace_kernels["metal"](bitmask, logits)
+    # # If Metal implementation is available, use it
+    # if "metal" in xgrammar.kernels.apply_token_bitmask_inplace_kernels:
+    #     return xgrammar.kernels.apply_token_bitmask_inplace_kernels["metal"](bitmask, logits)
 
     # Otherwise, fall back to the naive implementation
     vocab_size = logits.shape[-1]
@@ -47,6 +49,10 @@ def apply_token_bitmask_inplace_mlx(bitmask: torch.Tensor, logits: mx.array) -> 
     return logits + mx.array(logits_mask.numpy())
 
 
+g_bitmasks = []
+g_logits = []
+
+
 class XGrammarLogitsProcessor:
     def __init__(self, grammar: xgrammar.CompiledGrammar, max_rollback_tokens: int = 16):
         self.matcher = xgrammar.GrammarMatcher(grammar, max_rollback_tokens=max_rollback_tokens)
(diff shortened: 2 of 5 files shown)


wangkuiyi:
I’ve updated python/xgrammar/contrib/mlxlm.py, so we can now run the following command to compare:
1. Approach 3 – manually written Metal kernel compiled by `mlx.core.fast.metal_kernel`.
2. Approach 4 – `mx.compile`-accelerated expand-and-apply.
3. Approach 1 – naive expand-and-apply.

```shell
python python/xgrammar/contrib/mlxlm.py --model mlx-community/Qwen2.5-Coder-32B-Instruct-3bit
```

The benchmark results below are based on a real use case: running Qwen2.5 Coder 35B to generate a structured JSON response from a real prompt.

```
apply_bitmask_inplace_kernels[metal]: 0.0008109583053737879 seconds
apply_bitmask_compiled: 0.0036773292114958165 seconds
apply_token_bitmask_inplace_mlx: 0.31318166670389475 seconds

apply_bitmask_inplace_kernels[metal]: 0.0002715208800509572 seconds
apply_bitmask_compiled: 0.00022429581731557845 seconds
apply_token_bitmask_inplace_mlx: 0.31162573751062156 seconds

apply_bitmask_inplace_kernels[metal]: 0.0002623375039547682 seconds
apply_bitmask_compiled: 0.00022329168859869242 seconds
apply_token_bitmask_inplace_mlx: 0.3169941375032067 seconds
```

We ran three sets of comparisons. The first includes the compilation time of mx.compile, so you can ignore that one. The latter two show consistent trends:

1. Both the handwritten Metal kernel and the compiled kernel are approximately 1000× faster than the naive implementation.
2. The mx.compile version is slightly faster than the manually written Metal kernel—likely due to kernel fusion or better scheduling.

## Conclusion

We should adopt Approach 4. I will open another PR.

Ubospica:
#287

PR #289, opened 2025-04-02
Float range support for xgrammar

zanderjiang:
This PR adds support for handling float type values in json schemas, it can efficiently parse float ranges similar to GenerateRangeRegex. Floating point can be set manually to arbitrary precision as needed (assuming 6 as standard).


Ubospica:
@zanderjiang Thanks for your contribution!

PR #290, opened 2025-04-03
Disable slow C++ test cases

Ubospica:
This PR disables slow C++ test cases. These tests should ideally disable in a configurable manner. cc @DarkSharpness


DarkSharpness:
maybe we can try `ctest -E` to exclude useless tests

Ubospica:
@DarkSharpness Can that be achieved through a config file?

PR #294, opened 2025-04-08
Add util function testing._is_single_token_bitmask

Ubospica:
This PR adds a function _is_single_token_bitmask to check if the bitmask only allows one token.

### DIFF
--- a/tests/python/test_token_bitmask_operations.py
+++ b/tests/python/test_token_bitmask_operations.py
@@ -8,7 +8,11 @@ import pytest
 import torch
 
 import xgrammar as xgr
-from xgrammar.testing import _bool_mask_to_bitmask, _get_masked_tokens_from_bitmask
+from xgrammar.testing import (
+    _bool_mask_to_bitmask,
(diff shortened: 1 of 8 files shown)


tests/python/test_token_bitmask_operations.py
```diff
@@ -38,6 +42,27 @@ def test_get_masked_tokens_from_bitmask(token_mask_size: int, index: int):
...
+    bool_mask[batch_index, token_id] = True
+    bitmask = _bool_mask_to_bitmask(bool_mask)
+    assert _is_single_token_bitmask(bitmask, vocab_size, batch_index) == (True, token_id)
+    bool_mask[batch_index, token_id + 1] = True
+    bitmask = _bool_mask_to_bitmask(bool_mask)
+    assert _is_single_token_bitmask(bitmask, vocab_size, batch_index) == (False, -1)
+
+
+test_is_single_token_bitmask()
+exit()
```
zbowling, line 63: This looks like a mistake. It's breaking tests in 0.1.18

PR #305, opened 2025-04-22
[fix]fix the handling for '\?'.

Seven-Streams:
It's a minor bug. When handling escape characters, it can't handle `\?` correctly at present. This PR fixed the problem.


Ubospica (approved):
Thanks @Seven-Streams

PR #301, opened 2025-04-13
[Ci] Add a check in benchmark.yaml.

Seven-Streams:
At present, bennchmark.yaml will automatically run every day, and often annoy me with a "Run failed". Adding a check may be better.

PR #309, opened 2025-04-28
[CI Fix] Fix the check for benchmark.

Seven-Streams:
#301 I managed to solve the problem of CI(but failed). This PR should correctly fix it(I promise!).


Ubospica (approved):
LGTM!

PR #312, opened 2025-05-05
[Fix] matching error when encountering multi-byte UTF-8 encoded characters

Irfnfnkemed:
This PR fixed the matching error when encountering multi-byte UTF-8 encoded characters.

-  Issues previously existed: the matcher may unexpectedly reject the character when encountering multi-byte UTF-8 encoded characters. For example: `tags=[(beg="！！！", ... ), ...], triggers=["！",...]`.  For details see the test script.


Ubospica:
Thanks! @Irfnfnkemed

PR #321, opened 2025-05-26
[Feat] Enhance the json schema

Irfnfnkemed:
This PR enhances the json schema.
* Support `patternProperties` and `propertyNames` keywords in Object type. Currently, only the case where `additionalProperties=False` and both `properties` and `required` are empty is supported. Otherwise, potential conflicts may arise, which cannot be resolved at present. Also the situation of duplicate keys has not been resolved yet.
* Support `minProperties` and `maxProperties` keywords in Object type. The complexity of the rule will be related to $properties\\_size^2$ when `minProperties` or`maxProperties`  is set.
* Refactor the processing of the object schema. The process of parsing the initial schema and generating the EBNF has been decoupled.
* Add corresponding testpoints.


cpp/json_schema_converter.cc
```diff
@@ -2262,189 +2296,570 @@ std::string JSONSchemaConverter::GetPartialRuleForPropertiesAllOptional(
...
+    additional_suffix = "uneval";
+    additional_property = object_spec.unevaluated_properties_schema;
+  }
+  indentManager_->StartIndent();
+
+  if (object_spec.pattern_properties.size() > 0 ||
+      !object_spec.property_names.is<picojson::null>()) {
+    // Case 1: patternProperties or propertyNames is difined
+    // TODO: Here we only handle the case that additionalProperties=False
+    // TODO: The coexistence of properties, required, etc. has not been addressed yet,
```
Ubospica, line 2796: Support properties

cpp/json_schema_converter.cc
```diff
@@ -2262,189 +2296,570 @@ std::string JSONSchemaConverter::GetPartialRuleForPropertiesAllOptional(
...
+  }
   return res;
 }
 
-std::string JSONSchemaConverter::VisitObject(
-    const picojson::object& schema, const std::string& rule_name
+Result<JSONSchemaConverter::ObjectSpec> JSONSchemaConverter::ParseObjectSchema(
+    const picojson::object& schema
 ) {
   XGRAMMAR_CHECK(
```
Ubospica, line 2603: DCHECK

tests/python/test_json_schema_converter.py
```diff
@@ -1062,6 +1091,330 @@ def test_object_with_only_properties_keyword():
...
+            for instance, num_properties, have_additional in instances:
+                if num_properties < lower_bound or num_properties > upper_bound:
+                    is_accepted = False
+                else:
+                    is_accepted = True
+                check_schema_with_instance(
+                    schema, instance, is_accepted=is_accepted, any_whitespace=False
+                )
+
+
```
Ubospica, line 1417:
```
pytest.throws(RuntimeError, msg_regex): # required must be an array
   ...
```

PR #322, opened 2025-05-27
[Feature] Reject invalid tag dispatch grammar when compiling.

Seven-Streams:
Typically, in tag dispatch, a tag shouldn't be a prefix of one of the others. At present, such grammars will be accepted, even without warnings, which can lead to some confusing situations. This PR will reject such grammars when the users are trying to compile them.


Ubospica:
I think we shall further support such cases instead of rejecting it; also the error should be raised with parsing grammars. We can consider implementing this approach later. But thank you for your contribution!

PR #326, opened 2025-06-01
[Test] Add more tests to test fsm.

Seven-Streams:
This PR adds more tests to test the new fsm lib, and fixes some bugs in the lib.

### DIFF
--- a/cpp/fsm.cc
+++ b/cpp/fsm.cc
@@ -5,6 +5,7 @@
 #include "fsm.h"
 
 #include <algorithm>
+#include <bitset>
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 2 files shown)


cpp/fsm.cc
```diff
@@ -1293,7 +1256,7 @@ FSMWithStartEnd FSMWithStartEnd::MinimizeDFA() const {
 }
 
 FSMWithStartEnd FSMWithStartEnd::ToDFA() const {
-  FSMWithStartEnd dfa(0, start_, {}, true);
+  FSMWithStartEnd dfa(0, 0, {}, true);
```
Ubospica: Can you add explicit to 
```
FSM(int num_states = 0);
```
and change this to 
```suggestion
  FSMWithStartEnd dfa(FSM(0), 0, {}, true);
```

tests/cpp/test_fsm.cc
```diff
@@ -265,6 +265,17 @@ TEST(XGrammarFSMTest, FunctionTest) {
   fsm_wse = fsm_wse.SimplifyEpsilon();
   EXPECT_TRUE(fsm_wse.AcceptString(test_str));
   EXPECT_EQ(fsm_wse->NumStates(), 2);
+  std::cout << "--------- Function Test10 -----------" << std::endl;
```
Ubospica, line 268: We shall use structural tests in addition to the string acceptance tests. The test would be:
```
std::string printed = fsm.Print();
std::string expected = R"(.....)";
EXPECT_EQ(printed, expected);
```

It's not a high priority, but it would be good to do so later.

Ubospica:
Thanks @Seven-Streams !

PR #277, opened 2025-03-27
[Feature] Json Serialize Method

DarkSharpness:

PR #335, opened 2025-06-09
[Feature] JSON Serialize Method with C++ reflection

DarkSharpness:
This PR provides functionality similar to #277, but leverages C++ reflection to eliminate the need for manually defining serialization/deserialization rules for each class. It builds on top of #334, and should be merged after #334 is approved.

### What's new in this PR:

1. Introduced `AutoJSONSerialize` and `AutoJSONDeserialize` to enable reflection-based JSON serialization.
2. Refactored `tokenizer_info.cc`: moved the definition of `TokenizerInfo::Impl` into a separate file to support serialization.
3. Added Python bindings for JSON serialization (via `JSONSerializer`) and corresponding tests.


hnyls2002:
@DarkSharpness @Ubospica  Hi, I really suggest that all the serializable classes should implement a pickle trait like https://stackoverflow.com/questions/1939058/simple-example-of-use-of-setstate-and-getstate

Ahajha:
Driveby comment, are you aware of glaze? It's a reflection-based JSON de/serialization library. Maybe it could be useful here? https://github.com/stephenberry/glaze

Ubospica:
@Ahajha Thanks for the suggestion! Will check it out

Ubospica (approved):
LGTM. We shall consider the json library in future versions.

Ubospica:
This is a rather important feature. Thanks @DarkSharpness

PR #336, opened 2025-06-15
Convert Result to use unique_ptrs

Ahajha:
There seems to be no advantage of shared_ptr here - convert to unique_ptr to reduce overhead slightly.

This also removes the `const&` overload of `UnwrapErr`, since we can no longer copy the pointer, but in all use cases we can convert to using the `&&` overload instead.

### DIFF
--- a/cpp/fsm.cc
+++ b/cpp/fsm.cc
@@ -682,7 +682,7 @@ Result<FSMWithStartEnd> FSMWithStartEnd::Intersect(
     const FSMWithStartEnd& lhs, const FSMWithStartEnd& rhs, const int& num_of_states_limited
 ) {
   if (!lhs.IsLeaf() || !rhs.IsLeaf()) {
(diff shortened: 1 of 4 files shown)


Ubospica:
Thanks! We are currently refactoring the Result data structure. It will use a `std::variant<T, E> data_;` instead of the current 
```c++
std::optional<T> value_;
std::shared_ptr<Error> error_;
```
the change is part of https://github.com/Ubospica/xgrammar/pull/6/files#diff-5a94e96b2a1a18a5a568e2cf145ae734b8b9376b3474217628036c989caeeabf, and will be moved to the main repo soon. Plz check that out and feel free to leave comments!

Ahajha:
Ah nice! If that's the case, I think this PR isn't needed then?

Ubospica:
I think so. Thanks again!

PR #337, opened 2025-06-15
Make sentencepiece and tiktoken optional deps

Ahajha:
Sentencepiece has not had a new release in over a year, and users are running into easily fixable install issues like CMake minimum versions, missing headers, and missing wheels for newer python versions. See recent issues: https://github.com/google/sentencepiece/issues, like this one for example: https://github.com/google/sentencepiece/issues/1113

Tiktoken hasn't caused any particular issues, but I'm including it because it's used in the same capacity here.

Both dependencies are only actually used if users are, so it seems pointless to depend on them directly. This updates the support to silently skip over them if they are not installed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,8 +15,6 @@ keywords = ["machine learning", "inference"]
 requires-python = ">=3.8, <4"
 dependencies = [
   "pydantic",
-  "sentencepiece",
-  "tiktoken",
   "torch>=1.10.0",
   "transformers>=4.38.0",
   "triton; platform_system == 'Linux' and platform_machine == 'x86_64'",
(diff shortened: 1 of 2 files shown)


Ubospica (approved):
Thanks for the contribution @ahajha!

> Both dependencies are only actually used if users are, so it seems pointless to depend on them directly

This argument makes sense to me. May I ask if you have encountered any specific problems with the current dependencies?

Ahajha:
@Ubospica Yes, we (Modular) got this user report: https://github.com/modular/modular/issues/4509. We don't use the dependency directly.

Ahajha:
Dependencies in general, worth noting is triton and pytorch, but those are likely difficult to get rid of. We are (in an extremely simplified sense) competing with pytorch, so not depending on them is an ideal goal for us, though we use it in other places in our stack so I wouldn't sweat trying to get rid of it.

Issue #233, opened 2025-03-09
Avoid extra indirections in Python bindings


Ahajha:
This is a neat project! I have a personal interest in Python bindings, and I see some room for improvement that should give a performance bump and memory savings. View this issue as more of a planning doc for me to dump these ideas I have - they're all vaguely connected to the same core idea of reducing indirection.

In the current state of things, we have several levels of indirection when we want to call a method bound in C++. I will use `Grammar.union` as an interesting example, which in Python is defined as follows:
```python
grammar_handles = [grammar._handle for grammar in grammars]
return Grammar._create_from_handle(_core.Grammar.union(grammar_handles))
```
I mainly use this as an example of some low-hanging fruit, various other methods might be more complicated (and thus more difficult to improve), or less complicated (and thus have less headroom to improve). Though just about the entire API surface can have most of these improvements apply, if not all.

1. Call the method written in Python
2. Look up all of the underlying `_handle`s of the input
3. Create an intermediate list of these results
4. Call the underlying C++ function - this involves converting the intermediate list to a `std::vector<Grammar>`. The underlying objects are "copied" (which are really just `std::shared_ptr`s).
5. This then calls the underlying PIMPL function
6. The result is wrapped in a shared_ptr and returned as a C++ `Grammar`
7. The C++ result is wrapped in a Python object and returned to Python
8. The Python handle is then passed to `_create_from_handle` to create another Python object, which is returned as the final output.

I am proposing that this can be cut down to:
1. Call the underlying C++ function - This involves creating an intermediate vector of references to all of the `Grammar`s in the input list
2. The function (core logic) is called (no PIMPL)
3. The result is wrapped in a Python object, which is returned to Python

The memory layout is also worth looking at. Currently, to get from the top level Python `Grammar` object to the underlying C++ data, we need the following steps:
1. Access the `_handle` field of the Python object, which references another Python object
2. Look into the pybind11 internals of the object - by default this holds a `std::unique_ptr` to the underlying C++ object
3. Dereference the unique_ptr to get the C++ class
4. Dereference the shared_ptr in the C++ class to get to the PIMPL class, which contains the actual data.

I proposing that this can be cut down to:
1. Access the data directly in the Python object

I also propose that these changes can be made with minimal to no impact on the Python API, though with some modification to the C++ API, mainly with respect to ownership.

The main steps that are involved to achieve this are as follows (these can likely be done in any order, aside from the conversion to nanobind which should be done first)
[...]

PR #338, opened 2025-06-16
Bind fill_next_token_bitmask against nb::ndarray

Ahajha:
Nanobind has built-in support for the DLPack standard, which allows us to take anything adhering to the DLPack spec here rather than just PyTorch objects. See here: https://nanobind.readthedocs.io/en/latest/ndarray.html

This does add a dependency on numpy, but only for a single type, which does seem a little overkill. If I push on https://github.com/mlc-ai/xgrammar/issues/233 a little bit though, we should only need these definitions in the stubfiles, and will be unneeded for the `.py` files. So ideally this is a temporary dependency.

My goal is to reduce the dependency on PyTorch a bit by making the code more general. I don't know if (or even think) that we can remove it entirely, but this seems worthwhile to do.

The annotation on the `nb::ndarray` class will actually check at runtime that the parameter has the desired properties (in this case, it's on the CPU and is an int32_t), and will simply fail to call the function if not. This might be a slight breaking change in terms of what exception actually gets raised, but I think this is reasonable.

A few related changes:
- I've converted a few function signatures to take DLTensors by `const&` instead of `*`, didn't see a good reason for it.
- Added an anonymous namespace in `nanobind.cc` just to keep things hygenic.


cpp/nanobind/nanobind.cc
```diff
@@ -39,6 +42,32 @@ std::vector<std::string> CommonEncodedVocabType(
   return encoded_vocab_strs;
 }
 
+bool GrammarMatcher_FillNextTokenBitmask(
+    GrammarMatcher& matcher,
+    nb::ndarray<int32_t, nb::device::cpu> arr,
```
Ubospica: Can we accept any array here and report a customized error msg when the array is not on CPU? The msg can be
```
fill_next_token_bitmask requires the tensor to be on CPU, but got device ...
```

cpp/grammar_matcher.cc
```diff
@@ -270,7 +270,9 @@ class GrammarMatcher::Impl : public GrammarMatcherBase {
 
   bool AcceptString(const std::string& input_str, bool debug_print = false);
 
-  bool FillNextTokenBitmask(DLTensor* next_token_bitmask, int index, bool debug_print = false);
+  bool FillNextTokenBitmask(
+      const DLTensor& next_token_bitmask, int index, bool debug_print = false
```
Ubospica: The DLTensor* input mainly indicates the function will "modify" the content of the tensor. It's a convention in XGrammar that all functions modifying their input will pass by pointer. So I think we can keep DLTensor* here.

python/xgrammar/matcher.py
```diff
@@ -269,7 +270,7 @@ def accept_string(self, input_str: Union[str, bytes], *, debug_print: bool = Fal
         return self._handle.accept_string(input_str, debug_print)
 
     def fill_next_token_bitmask(
-        self, bitmask: torch.Tensor, index: int = 0, *, debug_print: bool = False
+        self, bitmask: ArrayLike, index: int = 0, *, debug_print: bool = False
```
Ubospica, line 285: What about
```suggestion
        self, bitmask: Union[torch.Tensor, ArrayLike], index: int = 0, *, debug_print: bool = False
```
I still wanna explicitly mention it can accept torch.Tensor because it's the default mask type.

Ubospica (commented):
Sorry, I just got time to review this. I have left some comments. It's generally good to support non-torch tensor types, like mlx, tensorflow, jax.array, mojo tensor, etc. 

Currently there are still relatively few users using non-torch tensors, but I do hope more users can leverage XGrammar to achieve structured generation in different platforms.

You’re also very welcome to support Mojo. If you’re interested, you can implement support for Mojo tensors in contrib/mojo.py.

Ubospica:
@Ahajha I have made the changes. Please check it out and you're welcome to make comments!

cpp/grammar_matcher.cc
```diff
@@ -274,9 +274,7 @@ class GrammarMatcher::Impl : public EarleyParser {
 
   bool AcceptString(const std::string& input_str, bool debug_print = false);
 
-  bool FillNextTokenBitmask(
-      const DLTensor& next_token_bitmask, int index, bool debug_print = false
-  );
+  bool FillNextTokenBitmask(DLTensor* next_token_bitmask, int index, bool debug_print = false);
```
Ahajha: Converting back to `DLTensor*` is fine with me :+1:

cpp/nanobind/nanobind.cc
```diff
@@ -21,9 +21,8 @@
 #include "python_methods.h"
 
 namespace nb = nanobind;
-using namespace xgrammar;
 
-namespace {
+namespace xgrammar {
```
Ahajha: This anonymous namespace was intentional - it probably doesn't make too much of a difference, but it marks any functions within it as compilation-unit-local, which could help optimize the code. It also can give warnings if anything in it isn't used. Since the functions are just used in the file, we can mark it as such.

Probably not a huge deal, but that was the motivation.

cpp/nanobind/nanobind.cc
```diff
@@ -43,13 +42,20 @@ std::vector<std::string> CommonEncodedVocabType(
...
-    GrammarMatcher& matcher,
-    nb::ndarray<int32_t, nb::device::cpu> arr,
-    int32_t index,
-    bool debug_print
+    GrammarMatcher& matcher, nb::ndarray<> arr, int32_t index, bool debug_print
 ) {
   if (arr.ndim() != 1 && arr.ndim() != 2) {
-    throw nb::type_error("token_bitmask tensor must be 1D or 2D");
+    throw std::runtime_error("token_bitmask tensor must be 1D or 2D");
+  }
```
Ahajha, line 49: This seems good to me!

Ahajha (commented):
Here's where I'm not sure about. I have a few  goals here that I should probably lay out. Though admittedly, we (Modular) might be switching away from `xgrammar` for some of these.

The whole reason we're even considering switching away is because of the PyTorch dependency. We're in a sense a PT competitor, and so having it as a dependency is weird. It's also a packaging nightmare - it's massive, and you have to opt-in to a CPU-only version which is smaller if that's all you use (which is the case for us currently). So our options here are either switch away, use xgrammar as a C++ library (in which case this PR doesn't really matter, we'd just make the requisite bindings changes in-house), or submit the requisite changes to remove PT as a dependency here, which this is the only function touching PT that we actually use (I don't think we even use the kernels AFAICT).

All that said, by annotating the type with `torch.tensor`, we keep the PT dependency around. I think there's a way we could mark it as annotation-only, but we'd still need to do a bit of work to remove the dep from the rest of the file. At the end of the day, for this to work for us, the PT dependency would have to be dropped, which probably the only way forwards is to push everything else into an extra, which unfortunately there's no (current) way to specify a default extra. I think we're of the mindset that it's a good project, just not a good fit for us.

The other goal is just a personal one which is moving more code to bindings, I'll respond to that comment on the issue shortly but the idea of using `ArrayLike` is that that is what nanobind autogenerates for ndarrays, so it would carry over nicely.

cpp/nanobind/nanobind.cc
Ubospica: Yes, I understand the motivation. But since we have enabled LTO, the difference between the two could be minor. So both is fine I think.

Ubospica:
> Here's where I'm not sure about. I have a few goals here that I should probably lay out. Though admittedly, we (Modular) might be switching away from `xgrammar` for some of these.
> 
> The whole reason we're even considering switching away is because of the PyTorch dependency. We're in a sense a PT competitor, and so having it as a dependency is weird. It's also a packaging nightmare - it's massive, and you have to opt-in to a CPU-only version which is smaller if that's all you use (which is the case for us currently). So our options here are either switch away, use xgrammar as a C++ library (in which case this PR doesn't really matter, we'd just make the requisite bindings changes in-house), or submit the requisite changes to remove PT as a dependency here, which this is the only function touching PT that we actually use (I don't think we even use the kernels AFAICT).
> 
> All that said, by annotating the type with `torch.tensor`, we keep the PT dependency around. I think there's a way we could mark it as annotation-only, but we'd still need to do a bit of work to remove the dep from the rest of the file. At the end of the day, for this to work for us, the PT dependency would have to be dropped, which probably the only way forwards is to push everything else into an extra, which unfortunately there's no (current) way to specify a default extra, so it would break users (you'd have to require something like `xgrammar[torch]`. I think we're of the mindset that it's a good project, just not a good fit for us.
> 
> The other goal is just a personal one which is moving more code to bindings, I'll respond to that comment on the issue shortly but the idea of using `ArrayLike` is that that is what nanobind autogenerates for ndarrays, so it would carry over nicely.

@Ahajha Thanks for the response! 

I think it’s reasonable to gradually separate out the Torch-related code (for example, placing it in contrib/torch.py). But several issues arise when we want to maintain backward compatibility:
1. allocate_token_bitmask still needs to return a torch.Tensor, and since it’s a public API, that means the codebase still needs to depend on Torch.
2. Regarding Python dependencies, one possible solution is to let xgrammar[all] install all dependencies, while xgrammar itself doesn’t depend on Torch. I’m just a bit concerned that this might break backward compatibility with pip install xgrammar.

I’d like to hear your thoughts on these two issues.

At the very least, moving Torch-related code to contrib/torch.py might make it easier for you to fork the code and remove torch dependency for your usecase.

As for this PR, I think we can change the annotation to ArrayLike. Can you also mention in the Python docstring that this parameter `accepts array-like objects such as torch.Tensor that supports the DLPack procotol`? Then we can go ahead and merge it.

Ubospica:
@Ahajha I think we can update to remove torch.Tensor in the type annotation, and merge this PR. I can do it tomorrow. If you have any further comments, feel free to share them!

python/xgrammar/matcher.py
```diff
@@ -281,7 +282,7 @@ def accept_string(self, input_str: Union[str, bytes], *, debug_print: bool = Fal
         return self._handle.accept_string(input_str, debug_print)
 
     def fill_next_token_bitmask(
-        self, bitmask: torch.Tensor, index: int = 0, *, debug_print: bool = False
+        self, bitmask: Union[torch.Tensor, ArrayLike], index: int = 0, *, debug_print: bool = False
```
Ubospica:
```suggestion
        self, bitmask: ArrayLike, index: int = 0, *, debug_print: bool = False
```

Ubospica (approved):

Ubospica:
@Ahajha Thank you for your contribution!

Ahajha:
@Ubospica Sorry about the lack of response. Yes that should be fine, thanks!

Ubospica:
> @Ubospica Sorry about the lack of response. Yes that should be fine, thanks!

No worries. Thank you!

PR #340, opened 2025-06-18
Add tiktoken and sentencepiece to test dependencies

Ahajha:
Not sure why it worked in presubmit but not on main, but there were some test failures due to tiktoken being missing. Readd that to the test dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,15 +29,14 @@ GitHub = "https://github.com/mlc-ai/xgrammar"
 
 [project.optional-dependencies]
 test = [
-  "pytest",
-  "protobuf",
   "huggingface-hub[cli]",
+  "protobuf",
+  "pytest",
   "sentencepiece",
   "tiktoken",
   # transformers==4.50.0 has error on MacOS.
   # https://github.com/huggingface/transformers/issues/36906
   "transformers<4.50.0; platform_system == 'Darwin'",
-  "tiktoken",
 ]
 
 [tool.scikit-build.metadata.version]


Seven-Streams:
It's because that when secrets.HF_TOKEN is not set, only part of unit tests will be run.

Ahajha:
In that case, why is the token only set for main? Is it for secret security?

Seven-Streams:
I think so. It seems like that secrets won't be passed to workflows triggered by a pull request from a fork in github.

Ahajha:
I know for a fact that they are, but I don't know if it's intentional or not to restrict it here.

Seven-Streams:
I'm not sure about that. @Ubospica

Seven-Streams:
Moreover, some unit tests depend on sentencepiece, maybe it also needs to be added again?

Ahajha:
I didn't see any specific failures pertaining to sentencepiece, but I think that's fine to re-add.

Ubospica (approved):
LGTM. Thanks!

Ubospica:
I think GitHub doesn't grant access to secrets for actions triggered by `pull_request`, mainly for security reasons. Let's check if the ci passes on main.

Ahajha:
Yea I learned recently about the secrets + pull requests thing, sorry for being so certainly wrong :sweat_smile:. My understanding now is that the secret is only available in this repo, so it only is available on PRs from branches from this repo.

Ubospica:
Yes indeed. It's like a tradeoff between security and convenience

PR #341, opened 2025-06-18
Generate and install stubfiles into the wheel

Ahajha:
This uses nanobind's builtin stub generator tool to generate stubfiles. These can be seen by generating a wheel locally, and these should help with IDE go-to-definitions, though they'll just point at a stub. We can add better type annotations, docs, and argument names to the compiled stubs later.

Working towards https://github.com/mlc-ai/xgrammar/issues/233


Ahajha:
Going to wait for https://github.com/wjakob/nanobind/pull/1082 - and I think I can actually do this in a better way, we can run this script at install time to just directly put the stubs in the wheel. Will need to give it some thought. Note for self: https://stackoverflow.com/questions/41254902/cmake-run-script-for-install-target

PR #352, opened 2025-07-02
Fix: fix wrong LRU cache implementation

DarkSharpness:
1. Fix the bug when an exception happens in creating an entry
2. Fix missing clear map when LRU is not enabled (unlimited case)

When `std::packaged_task` gets an exception inside the task, the associated `std::shared_future` will rethrow the exception on calling `get()` method. When this happens in LRU operations, this will lead to underlying cache inconsistency.

We fix this by catching all the possible exceptions (hopefully) when `get()` is called.


hnyls2002:
Thanks for fixing it.😭

DarkSharpness:
> Thanks for fixing it.😭

My bad sorry. I completely forgot about C++ exception in my first version of LRU cache 💩.

PR #359, opened 2025-07-11
Support incontiguous logits and bitmask.

hnyls2002:
This PR supports incontiguous logits and bitmasks when applying token bitmasks (The vocab stride still requires to be 1)


Ubospica (approved):
LGTM! We can add a test and merge it

Ubospica:
We can merge it for now and add tests later

PR #361, opened 2025-07-11
[Feat] Enhance the json schema

Irfnfnkemed:
This PR enhances the json schema.
* Support `patternProperties` and `propertyNames` keywords in Object type. Currently, only the case where `additionalProperties=False` and both `properties` and `required` are empty is supported. Otherwise, potential conflicts may arise, which cannot be resolved at present. Also the situation of duplicate keys has not been resolved yet.
* Support `minProperties` and `maxProperties` keywords in Object type. The complexity of the rule will be related to $properties\\_size^2$ when `minProperties` or`maxProperties`  is set.
* Refactor the processing of the object schema. The process of parsing the initial schema and generating the EBNF has been decoupled.
* Add corresponding testpoints.


Ubospica:
Thanks @Irfnfnkemed! The enhancement of JSON schema is very meaningful. Now we support most of the json schema features.

PR #368, opened 2025-07-18
[Feature] Add a new expression to represent repetition to speed up.

Seven-Streams:
Currently, if we try to compile a grammar with a expression like `[a-z]{2, 20000}`, it will cost a lot of time to compile the grammar. In this PR, we add a new type of expression `kRepeat` in the parser, which can significantly improve the efficiency in such cases.

### DIFF
--- a/cpp/grammar_parser.cc
+++ b/cpp/grammar_parser.cc
@@ -7,12 +7,16 @@
 
 #include <picojson.h>
 
+#include <cstddef>
+#include <cstdint>
 #include <variant>
+#include <vector>
 
 #include "grammar_builder.h"
 #include "grammar_impl.h"
 #include "support/encoding.h"
 #include "support/logging.h"
+#include "xgrammar/grammar.h"
 
 namespace xgrammar {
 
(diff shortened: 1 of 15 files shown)


cpp/grammar_parser.cc
```diff
@@ -757,57 +761,59 @@ int32_t EBNFParser::HandleQuestionQuantifier(int32_t grammar_expr_id) {
...
+  // Add the lookahead elements
+  std::vector<int32_t> lookahead_elements = elements;
+  if (elements.empty()) {
+    return builder_.AddEmptyStr();
   }
-  for (int64_t i = 0; i < upper - lower - 1; ++i) {
-    auto ref_to_next_rule = builder_.AddRuleRef(rest_rule_ids[i + 1]);
-    auto new_grammar_expr_id = builder_.AddChoices(
-        {builder_.AddEmptyStr(), builder_.AddSequence({grammar_expr_id, ref_to_next_rule})}
+  for (size_t i = 0; i < elements.size() - 1; i++) {
```
Ubospica:
```suggestion
  for (int64_t i = 0; i < static_cast<int64_t>(elements.size()) - 1; i++) {
```
Avoid overflow

PR #369, opened 2025-07-20
Fix: fix wrong DEBUG boundary check

DarkSharpness:
When compiled with `CMAKE_BUILD_TYPE Debug`, we found that some of the tests crashed.

Try this in main branch and it will lead to some crashs:

```
pytest -k test_empty_
```

I would suggest turning on `XGRAMMAR_DCHECK` in ci to detect more bugs like this.

One more suggestion. I don't think magic numbers like `-1` is good enough. Maybe we can rewrite with `std::optional<int>` for better readability in future.


Ubospica (approved):
LGTM. I agree we should avoid using -1.

PR #370, opened 2025-07-20
[Feature] support internal check option in cmake

DarkSharpness:
In old main branch, whether debug check is enabled is controlled by `NDEDBUG`. Now, we introduce a new cmake option `XGRAMMAR_ENABLE_INTERNAL_CHECK` to control whether we enable internal check. This will allow us to perform some boundary checks even in release mode, which is helpful in ci.

We also update the github ci and enable internal check in ci in this PR (but this will lead to ci failure currently, see related bugs to debug check: #369)


Ubospica (approved):
LGTM. Thanks

PR #371, opened 2025-07-20
[Fix]fix behaviors and boundary check in earley parser.

Seven-Streams:
Like #369, this PR fixes wrong boundary check in earley parser. Moreover, it supports fsms in a more general way.
This PR also fixed some latent issues:
- Support epsilon edges in fsms.
- Disable right recursion optimization on fsms.
- Fix the boundary check. 
- Fix the parser when a fsm is referred by other rules.


Ubospica (approved):
Can you further mention: a) why there is a problem previously, b) what has been changed in this PR?

PR #372, opened 2025-07-22
[Fix] update the version of urllib3 in requirements.

Seven-Streams:


Ubospica:
Please double check if the docs can be built locally, and we can merge it

Seven-Streams:
Yes, the site can be sucessfully built locally.

PR #374, opened 2025-07-22
[Fix]Fix behaviors and boundary check in earley parser.

Seven-Streams:
This PR fixes wrong boundary check in earley parser. Moreover, it supports fsms in a more general way.
This PR also fixed some latent issues:
- Support epsilon edges in fsms.
- Disable right recursion optimization on fsms.
- Fix several boundary check.
- Fix the parser when a fsm is referred by other rules.


tests/cpp/test_fsm.cc
```diff
@@ -413,7 +413,7 @@ TEST(XGrammarFSMTest, EfficiencyTest) {
   time_end = std::chrono::high_resolution_clock::now();
   duration = std::chrono::duration_cast<std::chrono::milliseconds>(time_end - time_start);
   std::cout << "Time taken to minimize DFA: " << duration.count() << " ms" << std::endl;
-  EXPECT_EQ(fsm_wse->NumStates(), 111);
+  // EXPECT_EQ(fsm_wse->NumStates(), 111);
```
Ubospica: Why is this line commented?

Ubospica (approved):
In general LGTM. Is it possible to add regressions to this PR?

tests/cpp/test_fsm.cc
Seven-Streams: It's commented by mistake.

Ubospica (approved):
Thanks!

PR #387, opened 2025-08-08
[Feature] Provide a new function to convert xml style function calling to grammars.

Seven-Streams:
Consider we need to call a function:
```
LookUp(name: str, age: int)
```
And we need to require llms to output something like this:
```
<function=LookUp>
<parameter=name>Bob<\parameter>
<parameter=age>20<\parameter>
<\function>
```
Here, sometimes llms will output something invalid: invalid names of parameters, invalid type of parameters, etc.
This PR provides a new function, which can help constrain the output of the parameters:
```
def from_function_call(
        args_names: List[str], args_types: List[str], function_type: int = 0
    ) -> "Grammar":
```
`function_type` is reserved for future use.
`args_names` and `args_types` are Lists, which contains the names of the parameters, and the corresponding types of the parameters.
There are  several types will be dealt with differently:
- `int`, `uint`, `long`, `short`, `unsign`, `float`, `num`, or the types starting with theses words, will constrain the llm output a number.
- `boolean`, `bool`, `binary`, or the types starting with the words, will constrain the llm output `true` or `false`.
- `object`, `dict`, or the types starting with the words, will constrain the llm output a json object.
- `string`, `str`, `char`, `varchar`, `enum`, `text`, or the types starting with the words or any other types except those mentioned above, will constrain the llm output a xml-style string.

Signed-off-by: Yuchuan <blemiade_qinchuan@sjtu.edu.cn>


cpp/json_schema_converter.h
```diff
@@ -27,17 +32,22 @@ namespace xgrammar {
...
  */
+
 std::string JSONSchemaToEBNF(
     const std::string& schema,
     bool any_whitespace = true,
     std::optional<int> indent = std::nullopt,
     std::optional<std::pair<std::string, std::string>> separators = std::nullopt,
-    bool strict_mode = true
+    bool strict_mode = true,
+    StringEscapeType string_escape_type = StringEscapeType::kJSON
```
Ubospica: About API, let's use `enum class JSONFormat : int {kJSON=0, kXMLQwen=1};`

cpp/function_calling_converter.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "function_calling_converter.h"
+
+#include <optional>
+
+#include "json_schema_converter.h"
+#include "support/logging.h"
+
+namespace xgrammar {
+
+std::string QwenXMLToolCallingToEBNF(const std::string& schema) {
```
Ubospica: Put it in testing, `_QwenXMLToolCallingToEBNF`

cpp/json_schema_converter.cc
```diff
@@ -1966,7 +2073,19 @@ std::string JSONSchemaConverter::VisitString(
                              (max_length == -1 ? "" : std::to_string(max_length)) + "}";
     return "\"\\\"\" " + std::string("[^\"\\\\\\r\\n]") + range_part + " \"\\\"\"";
   }
-  return "[\"] " + kBasicStringSub;
+  switch (string_escape_type) {
```
Ubospica: About the architecture of JSONSchemaConverter, can we do this:
1. Some states are global. That can be stored in JSONSchemaConverter's members.
2. some states are local. We should store them into "Spec" classes, like StringSpec
3. In conversion, we convert according to states in 1 and 2
I feel the string format is 2 because it only affects the outermost layer. The parameter format is too.

Seven-Streams: Let me think for a while.

cpp/json_schema_converter.cc
```diff
@@ -458,6 +480,18 @@ class JSONSchemaConverter {
       const int max_properties
   );
 
+  /*! \brief Indicate the layer of the currrent object, used to determine the JSON format. */
+  class NestedLayerHelper {
```
Ubospica: We don't need this

cpp/json_schema_converter.cc
```diff
@@ -2790,18 +2887,268 @@ Result<JSONSchemaConverter::ObjectSpec, SchemaError> JSONSchemaConverter::ParseO
   });
 }
 
+Result<JSONSchemaConverter::StringSpec, SchemaError> JSONSchemaConverter::ParseStringSchema(
+    const picojson::object& schema
+) {
+  XGRAMMAR_DCHECK((schema.count("type") && schema.at("type").get<std::string>() == "string"));
```
Ubospica, line 2868: Move major logic to VisitString

cpp/json_schema_converter.cc
```diff
@@ -2790,18 +2887,268 @@ Result<JSONSchemaConverter::ObjectSpec, SchemaError> JSONSchemaConverter::ParseO
...
+      XGRAMMAR_LOG(FATAL) << "Unsupported string escape type: " << static_cast<int>(json_format);
+    }
+  }
+}
+
 std::string JSONSchemaConverter::VisitObject(
     const picojson::object& schema, const std::string& rule_name
 ) {
   // Parse the object schema
+  NestedLayerHelper helper;
```
Ubospica: ObjectSpec{JSONFormat format;};

// In converter.Convert
if (kXML) {
    // Do checking: schema must be object
         return SchemaError;
    return VisitXMLObject;
}
else {
  // Do normal conversion process
}

// In VisitXMLObject
ObjectSpec ... = parse();
spec.format = kXML;
Do conversion to bnf logic...

tests/python/test_function_calling_converter.py
```diff
@@ -0,0 +1,290 @@
+import sys
+
+import pytest
+
+from xgrammar.testing import _is_grammar_accept_string, _qwen_xml_tool_calling_to_ebnf
+
+test_string_schema_input_str_accepted = (
```
Ubospica, line 7: Add a test for format

Ubospica (approved):

PR #388, opened 2025-08-08
[Fix] Fix dependency on macos-13.

Seven-Streams:
This PR constrains the version of `safetensors` on macos-13. This addresses the compatibility between `safetensors` and `torch`.

ref:[https://github.com/ggml-org/llama.cpp/pull/15134](https://github.com/ggml-org/llama.cpp/pull/15134).
ref:[https://github.com/nvidia-holoscan/holohub/issues/1061](https://github.com/nvidia-holoscan/holohub/issues/1061).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dependencies = [
   "ninja",
   "numpy",
   "typing-extensions>=4.9.0",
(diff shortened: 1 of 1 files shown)


Seven-Streams:
It seems that they fixed the problem, thus the PR is closed.

PR #390, opened 2025-08-09
Fix apply_bitmask logit for both CPU and triton versions when shape and stride doesn't match

Jialin:
# Motivation
In vLLM, in order to address [Issue #19493](https://github.com/vllm-project/vllm/issues/19493), [PR #19565](https://github.com/vllm-project/vllm/pull/19565) served as a workaround to mitigate the issue by switching to use torch.compile version instead of triton version.

# Root cause
With some investigation, we found that the error would happen when logits.stride()[0] != logits.shape[-1].

# Changes
- Add a unit test to reproduce the errors
- Fix CPU and Triton-version apply_grammar_bitmask for this scenario

# Tests
We've verified
- New unit tests would failed on trunk
- New unit tests passed with the change (and no new failure, however, there're a lot of failing tests in trunk, 47 failed in tests/python/test_token_bitmask_operations.py)

# Followup
We will follow up with a benchmark comparison to see if we should bring back triton-version or use the new cuda version on vLLM side.

### DIFF
--- a/cpp/grammar_matcher.cc
+++ b/cpp/grammar_matcher.cc
@@ -113,10 +113,12 @@ void ApplyTokenBitmaskInplaceCPU(
       logits->ndim == 2
           ? std::make_pair(static_cast<int>(logits->shape[0]), static_cast<int>(logits->shape[1]))
           : std::make_pair(1, static_cast<int>(logits->shape[0]));
+  int logits_stride0 = logits->strides[0];
   std::pair<int, int> bitmask_shape =
       bitmask.ndim == 2
           ? std::make_pair(static_cast<int>(bitmask.shape[0]), static_cast<int>(bitmask.shape[1]))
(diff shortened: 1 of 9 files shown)


Ubospica (approved):
Thank you so much for your contribution!

strides are not considered previously. It's good that this PR fixes this issue.

In the future we will further use nanobind's ArrayList to simplify the logic of passing torch tensor to C++. See https://github.com/mlc-ai/xgrammar/blob/main/cpp/nanobind/nanobind.cc#L45. This can avoid the complex logic of passing shape, stride, and dtype, etc. separately to C++.

Jialin:
@Ubospica The precommit failure is addressed. And also we tried to fix cuda implementation in https://github.com/mlc-ai/xgrammar/pull/394

Ubospica:
@Jialin Thank you so much for your contribution

Issue #392, opened 2025-08-10
Multi-byte unicode characters in StructuralTagItem `end` / `begin` fields causes a crash


bdubayah:
## Issue Description

The `AddByteString` implementation in `GrammarBuilder` treats UTF-8 strings as raw byte sequences, causing multi-byte characters to be split into separate bytes.

## Reproduction

The following test case demonstrates the issue:

```python
import xgrammar
from transformers import AutoTokenizer


def test_xgrammar():
    model = "deepseek-ai/DeepSeek-V3-0324"
    tokenizer = AutoTokenizer.from_pretrained(model)
    tokenizer_info = xgrammar.TokenizerInfo.from_huggingface(tokenizer)
    compiler = xgrammar.GrammarCompiler(tokenizer_info)
    structures = [
        xgrammar.StructuralTagItem(
            begin="<｜tool▁calls▁begin｜>",
            schema=...,  # Any schema will work here
            end="<｜tool▁calls▁end｜>",
        )
    ]
    triggers = ["<｜tool▁calls▁begin｜>"]
    compiled_grammar = compiler.compile_structural_tag(structures, triggers)

```

This test produces the following error:

```
tests/unit/test_xgrammar.py Fatal Python error: Bus error
```

I tested this off of main. Let me know if there are any other details I can provide. Thanks so much for all of your work on xgrammar!

Seven-Streams:
@bdubayah Thanks for the bug reporting! The bug will be fixed in #396.

Seven-Streams:
The issue is resolved in #396 , and will be closed for now. Feel free to reopen it if you have further problems!

bdubayah:
Thanks so much for the quick fix!

PR #396, opened 2025-08-13
[Fix] Fix Multi-byte unicode characters in StructuralTagItem.

Seven-Streams:
As reported in #392 , Multi-byte characters in StructuralTagItem `end` / `begin` fields causes a crash. It is caused by the incorrect processing of multi-byte characters in building grammar from StructuralTagItem, and this mistake leads to the crash in grammar compiler. This PR fixes the problem.


Ubospica (approved):
LGTM. Thanks

Ubospica:
Plz fix conflict

Seven-Streams:
Fixed.

PR #409, opened 2025-09-02
[Fix] Fix regex converter when the property's pattern is empty.

Seven-Streams:
Currently, when we enconter such schema:
```
{
  "type": "object",
  "properties": {
    "_links": {
      "type": "object",
      "patternProperties": {
        "": {
          "type": "object",
          "properties": {
            "href": {
              "type": "string"
            }
          }
        }
      }
    }
  }
}
```
This will lead to a segmentation fault. It is caused by the lack of handling with the empty regex when `RegexConverter` is constructed. This PR fixed it.

Signed-off-by: Yuchuan <blemiade_qinchuan@sjtu.edu.cn>

### DIFF
--- a/cpp/regex_converter.cc
+++ b/cpp/regex_converter.cc
@@ -293,6 +293,7 @@ std::string RegexConverter::Convert() {
   start_ = regex_codepoints_.data();
   current_ = start_;
   end_ = start_ + regex_codepoints_.size() - 1;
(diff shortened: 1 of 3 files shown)


Seven-Streams:
> Plz add a test that uses empty regex to accept an empty str

Added

Ubospica (approved):
Thanks

PR #413, opened 2025-09-04
Update README.md to announce that OpenVINO GenAI uses XGrammar

apaniukov:
[OpenVINO GenAI](https://github.com/openvinotoolkit/openvino.genai) 2025.3.0 version adds Structured Output support with XGrammar backend.

[Release notes](https://docs.openvino.ai/2025/about-openvino/release-notes-openvino.html)

Samples:
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/structured_output_generation.py
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/structural_tags_generation.py
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/compound_grammar_generation.py
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/text_generation/structured_output_generation.cpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
 </div>
 
 ## News
+- [2025/09] XGrammar has been officially integrated into [OpenVINO GenAI](https://github.com/openvinotoolkit/openvino.genai)
 - [2025/02] XGrammar has been officially integrated into [Modular's MAX](https://docs.modular.com/max/serve/structured-output)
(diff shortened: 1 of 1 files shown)


Ubospica (approved):
Thanks! Welcome to use XGrammar, and we also welcome you to share more feedback.

PR #414, opened 2025-09-04
Json schema generation, limit the number of whitespaces

### LINKED ISSUE #412: During json schema convertion, limit the number of WS characters
Hello,

Example issue where it's stuck in an infinite generation loop : 

https://github.com/vllm-project/vllm/issues/21026

Image

For example like they do in llama.cpp https://github.com/ggml-org/llama.cpp/blob/master/grammars/json.gbnf#L25

ExtReMLapin:
Fixes #412


ExtReMLapin:
cc 

@Ubospica 

@Seven-Streams

Ubospica:
x@ExtReMLapin Thanks for the contribution! This is definitely useful. Could you fix the tests?

---

If you have extra time, instead of hardcoding the max number of whitespaces, you could consider making this a parameter of the JSON schema converter:
```
Grammar.from_json_schema(..., max_whitespaces: Optional[int] = None)
GrammarCompiler.compile_json_schema(..., max_whitespaces: Optional[int] = None)
```
And use this in the json schema converter.

With this, you don't need to fix all the tests, but only need to add a test case where `max_whitespaces != None`

ExtReMLapin:
For the tests, I went ooga booga and just kept replacing all the others whitespaces `*` occurences so it's easier to quickly fix all tests.


I still have 3 failing tests that kinda confuses me as they don't fail on main branch


For example : `tests/python/test_grammar_matcher_basic.py:177: AssertionError`

image



--------------------------------


Not going to lie, I understand it's better to add a param for quantity or param for direct grammar pattern, 100% agree.

At the office we're really missing time, I'm just here to fix an issue we have, I really don't have the bandwidth to go deeper contribution on the repositories we push changes to.

Seven-Streams:
@ExtReMLapin Thanks for the report and your contribution! I've opened a new pull request on your branch, and you can check if it fixes your issue. Moreover, if you have some free time, you can also add more tests for it.

ExtReMLapin:
Merged, your name should be on the PR, not mine :v 

At the end of the day it's fixed and that's all that matters imo !

Ubospica (approved):
@ExtReMLapin @Seven-Streams Thank you so much for your contribution! The next version will have this fix.

PR #417, opened 2025-09-08
[Refac] Refactor the bahavior of jsonschema converter to speed up.

Seven-Streams:
Notice that there will be some identical string structures in the same schema. For example, `{"type":"string", "maxLength": 1024}` can be contained in the schema a lot of times. This PR build a single rule for each identical string specs, and this can improve the efficiency in some cases.


cpp/json_schema_converter.cc
```diff
@@ -3172,6 +3218,7 @@ Result<JSONSchemaConverter::StringSpec, SchemaError> JSONSchemaConverter::ParseS
       return ResultOk(string_spec);
     }
     default: {
+      XGRAMMAR_UNREACHABLE();
       XGRAMMAR_LOG(FATAL) << "Unsupported JSON Format type: " << static_cast<int>(json_format);
```
Ubospica, line 3221: Swap the two lines

Issue #418, opened 2025-09-08
[Bug] Chinese enum values broken in v0.1.23 - UTF-8 bytes incorrectly interpreted as Latin-1 characters


kongjiellx:
## Description
Starting from version 0.1.23, xgrammar incorrectly handles Chinese characters in JSON schema enum values. The UTF-8 byte sequences are being interpreted as individual Latin-1 characters instead of proper Unicode code points.

### Expected Behavior
Chinese enum values should be encoded as proper Unicode escape sequences. For example:
- "在线" should be represented as `\u5728\u7ebf`
- "北京" should be represented as `\u5317\u4eac`

### Actual Behavior (v0.1.23)
UTF-8 bytes are incorrectly interpreted as Latin-1 characters:
- "在线" (UTF-8: `0xe5 0x9c 0xa8 0xe7 0xba 0xbf`) becomes `\u00e5\u009c\u00a8\u00e7\u00ba\u00bf`
- "北京" (UTF-8: `0xe5 0x8c 0x97 0xe4 0xba 0xac`) becomes `\u00e5\u008c\u0097\u00e4\u00ba\u00ac`

This causes the grammar matcher to reject valid Chinese tokens generated by the model.

### Minimal Example
```python
import xgrammar as xgr
from transformers import AutoTokenizer, AutoConfig
import json

# Setup
model_id = "Qwen/Qwen3-8B"  # or any model with Chinese support
tokenizer = AutoTokenizer.from_pretrained(model_id)
config = AutoConfig.from_pretrained(model_id)
full_vocab_size = config.vocab_size
tokenizer_info = xgr.TokenizerInfo.from_huggingface(tokenizer, vocab_size=full_vocab_size)

compiler = xgr.GrammarCompiler(tokenizer_info, max_threads=8)

# Schema with Chinese enum
schema = {
    'type': 'object',
    'properties': {
        'activity': {
            'type': 'string',
            'enum': ['在线', '北京']
        }
    }
}

compiled_grammar = compiler.compile_json_schema(json.dumps(schema))
print(compiled_grammar.grammar)
# In v0.1.23, this shows: \u00e5\u009c\u00a8\u00e7\u00ba\u00bf (incorrect)
# Expected: \u5728\u7ebf (correct)

matcher = xgr.GrammarMatcher(compiled_grammar, terminate_without_stop_token=True)

# Test Chinese token acceptance
test_json = '{"activity": "在线"}'
[...]

PR #380, opened 2025-07-27
[Refactor] JSON serializer and MemorySize 

DarkSharpness:
This is just a rebased version of #356.

PR #419, opened 2025-09-08
[Fix] Fix the matching for utf-8 characters in enums and const.

Seven-Streams:
As reported in #418, utf-8 characters will be interpreted incorrectly in `enum` and `const`. It is caused by #380 updated the serialization method, which is used during jsonschema conversion. This PR fixes the bug by refactoring the serialization method, and rollback the `picojson.h`

### DIFF
--- a/tests/cpp/test_serialization.cc
+++ b/tests/cpp/test_serialization.cc
@@ -253,7 +253,7 @@ TEST(XGrammarSerializationTest, TestString) {
   {
     std::string value = "\xC3\x28";
     auto json_value = AutoSerializeJSON(value);
-    ASSERT_EQ(json_value, "\"\u00c3(\"");
+    ASSERT_EQ(json_value, u8"\"\u00c3(\"");
 
     std::string deserialized;
     auto error = AutoDeserializeJSON(&deserialized, json_value);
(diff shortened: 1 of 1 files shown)


tests/cpp/test_serialization.cc
```diff
@@ -253,7 +253,7 @@ TEST(XGrammarSerializationTest, TestString) {
   {
     std::string value = "\xC3\x28";
     auto json_value = AutoSerializeJSON(value);
-    ASSERT_EQ(json_value, "\"\\u00c3(\"");
+    ASSERT_EQ(json_value, "\"\xC3\x83(\"");
```
Ubospica:
```suggestion
    ASSERT_EQ(json_value, "\"\u00c3(\"");
```

tests/cpp/test_serialization.cc
```diff
@@ -264,7 +264,7 @@ TEST(XGrammarSerializationTest, TestString) {
     std::string value = "我";
     auto json_value = AutoSerializeJSON(value);
     std::cout << json_value << std::endl;
-    ASSERT_EQ(json_value, "\"\\u00e6\\u0088\\u0091\"");
+    ASSERT_EQ(json_value, "\"\xC3\xA6\xC2\x88\xC2\x91\"");
```
Ubospica:
```suggestion
    ASSERT_EQ(json_value, "\"\u00e6\u0088\u0091\"");
```

Ubospica (approved):
LGTM

Issue #367, opened 2025-07-17
Python Grammar EBNF


paolovic:
Hello,

do you have some python grammar that I can feed into xgrammar as ebnf string to constrain my llm to syntactical correct python?
Or is there another way to use Xgrammar for constraining my LLM to correct python?
Similar to [SynCode](https://github.com/structuredllm/syncode/blob/main/syncode/parsers/grammars/python.lark) but for XGrammar usage.

Ubospica:
We have plan to support Python. Please stay tuned!

PR #422, opened 2025-09-09
[Feature] Enhance structural tags by providing xml-style tags.

Seven-Streams:
This PR is based on #387 and #420. This PR provides a new schema, which allows users to constrain the output with Qwen's xml-style function callings' parameters. For example:
```
xml_style = {
    "type": "qwen_xml_parameter",
    "json_schema": {
        "type": "object",
        "required": ["name", "age"],
    },
},
structural_tag = {"type": "structural_tag", "format": xml_style}
stag_ebnf = xgr.Grammar.from_structural_tag(structural_tag)
```
Then, we can use `stag_ebnf` to constrain the outputs. For example, `"<parameter=name>Bob</parameter><parameter=age>\t100\n</parameter>` is a valid output.

### DIFF
--- a/python/xgrammar/structural_tag.py
+++ b/python/xgrammar/structural_tag.py
@@ -74,14 +74,14 @@ class TagFormat(BaseModel):
     """The end tag."""
 
 
-class QwenXmlFormat(BaseModel):
+class QwenXMLParameterFormat(BaseModel):
     """A format that matches Qwen XML function calls."""
 
-    type: Literal["qwen_xml"] = "qwen_xml"
+    type: Literal["qwen_xml_parameter"] = "qwen_xml_parameter"
     """The type of the format."""
 
-    parameter_schema: Union[bool, Dict[str, Any]]
-    """The JSON schema for the parameters of the function call."""
+    json_schema: Union[bool, Dict[str, Any]]
+    """The JSON schema for the parameters of the function calling."""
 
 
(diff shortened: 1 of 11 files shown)


python/xgrammar/structural_tag.py
```diff
@@ -74,6 +74,16 @@ class TagFormat(BaseModel):
     """The end tag."""
 
 
+class QwenXmlFormat(BaseModel):
+    """A format that matches Qwen XML function calls."""
```
Ubospica: Add examples for format

Ubospica (approved):
Thanks!

Please also update the docs in docs/tutorials/structural_tag.md. Remember to add an example to illustrate the format, and mention this format is designed for the parameter format of qwen3-coder

Also plz rename to QwenXMLParameterFormat and "qwen_xml_parameter"

The format should be

```
        {
            "type": "qwen_xml_parameter",
            "json_schema": {
                "type": "object",
                "required": ["name", "age"],
            },
        },
```

Swipe4057:
Hi, maybe you know how to make a structured output so that the dictionary key value contains python code?

Seven-Streams:
> Hi, maybe you know how to make a structured output so that the dictionary key value contains python code?
@Swipe4057 #367 has requested the feature to generate python code. We plan to support it in the future.

PR #426, opened 2025-09-11
[Feature] Add a function to convert bitmasks to bool masks for development usage.

Seven-Streams:
This PR provides a function to convert a  bitmask to bool mask, the API is:
```
def _bitmask_to_bool_mask(bit_mask: torch.Tensor, vocab_size: Optional[int] = None) -> torch.Tensor:
```

### DIFF
--- a/tests/python/test_token_bitmask_operations.py
+++ b/tests/python/test_token_bitmask_operations.py
@@ -372,9 +372,17 @@ def test_bitmask_to_boolmask():
     assert torch.equal(bool_mask_50, expected_50)
 
 
-def test_bool_mask_bitmask_roundtrip():
-    batch_size = 4
-    vocab_size = 1000
+batch__size__vocab__size = [
+    (4, 1000),
+    (1, 1024),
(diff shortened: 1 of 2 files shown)


tests/python/test_token_bitmask_operations.py
```diff
@@ -357,5 +358,28 @@ def test_apply_token_bitmask_inplace_indices(
...
+    expected_50 = expected[:, :50]
+    assert torch.equal(bool_mask_50, expected_50)
+
+
+def test_bool_mask_bitmask_roundtrip():
+    batch_size = 4
+    vocab_size = 1000
+    bool_mask = torch.randint(0, 2, (batch_size, vocab_size), dtype=torch.bool)
+    bitmask = _bool_mask_to_bitmask(bool_mask)
+    bool_mask_converted = _bitmask_to_bool_mask(bitmask, vocab_size=vocab_size)
```
Ubospica, line 388: Plz add a test for vocab size not a multiple of 32

PR #429, opened 2025-09-12
[Fix] Apply the GrammarNormalizer to grammars generated from stags.

Seven-Streams:
This PR applies `GrammarNormalizer` to grammars generated from stags to improve the efficiency at runtime.

### DIFF
--- a/tests/python/test_grammar_matcher_structural_tag.py
+++ b/tests/python/test_grammar_matcher_structural_tag.py
@@ -98,7 +98,6 @@ def test_structural_tag():
     triggers = ["<function=f", "<function=g"]
 
     grammar = xgr.Grammar.from_structural_tag(tags, triggers)
-    print(grammar)
     assert str(grammar) == expected_grammar_test_structural_tag
 
     accepted_inputs = [


Ubospica (commented):
Have you tested the speed of compiling grammars (before & after on test cases, only reporting the slowest ones (like the ones >10ms))

Seven-Streams:
There is not an obvious difference between the before one and this one. About 87.6ms -> 82.7ms(16 threads, Ultra-7 155H). It seems to be a tolerable error.

Seven-Streams:
There is not an obvious difference between the before one and this one. About 87.6ms -> 82.7ms(16 threads, Ultra-7 155H). It seems to be a tolerable error.

tests/python/test_grammar_matcher_structural_tag.py
```diff
@@ -98,7 +98,7 @@ class Schema2(BaseModel):
     triggers = ["<function=f", "<function=g"]
 
     grammar = xgr.Grammar.from_structural_tag(tags, triggers)
-
+    print(grammar)
```
Ubospica: remove this

Ubospica (approved):

PR #431, opened 2025-09-14
[Optim] Optimize the efficiency of TagDispatch expressions.

Seven-Streams:
This PR optimized the efficiency of TagDispatch expression. At the initial state, only the tokens contain tags or stop strings are possible to be constrained. Otherwise, they are definitely accepted. This PR takes advantage of it, and brings more efficiency improvement.

### DIFF
--- a/cpp/grammar_compiler.cc
+++ b/cpp/grammar_compiler.cc
@@ -38,15 +38,14 @@ class GrammarMatcherForTokenMaskCache : public EarleyParser {
   GrammarMatcherForTokenMaskCache(
       const Grammar& grammar,
       const ParserState& init_state,
-      const std::unordered_map<int32_t, DynamicBitset>*
-          definite_accepted_in_tagdispatch_since_second_char_cache,
+      const std::unordered_map<int32_t, DynamicBitset>&
+          tag_dispatch_rule_id_to_second_slicing_bitset,
       const bool& need_expand = true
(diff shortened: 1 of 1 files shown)


cpp/grammar_compiler.cc
```diff
@@ -87,6 +98,10 @@ class GrammarMatcherForTokenMaskCache : public EarleyParser {
   // The initial state of the parser.
   ParserState initial_state;
 
+  // The definite accepted bitset in tag dispatch rules since the second character.
+  const std::unordered_map<int32_t, DynamicBitset>*
```
Ubospica: add more detailed docs

cpp/grammar_compiler.cc
```diff
@@ -606,7 +717,9 @@ CompiledGrammar GrammarCompilerNoCache::MultiThreadCompileGrammar(Grammar gramma
   }
 
   auto add_adaptive_token_mask = [&](const ParserState& state, bool is_root_rule) {
-    auto grammar_matcher = GrammarMatcherForTokenMaskCache(grammar, state, false);
+    auto grammar_matcher = GrammarMatcherForTokenMaskCache(
+        grammar, state, &definite_accepted_token_ids_since_second_char_cache_, false
```
Ubospica: use const& instead of const*

PR #433, opened 2025-09-16
[Refac] Refactor repetition expressions to educe the uncertainty.

Seven-Streams:
This PR refactors the processing of the repetition expression to reduce the uncertainty. For example, regexs like `[a-z]{0, 1000}`, whose minimum repetition time is small, will be handled better, and the uncertainty will be reduced.


Ubospica:
image
To update on the performance. Left: after, right: before. TPOT for large repetition grammars is significantly reduced.

PR #437, opened 2025-09-19
[Refac] Refactor the matcher for better efficiency.

Seven-Streams:
This PR refactors the logic of the matcher for better efficiency in long tail cases. At present, we'll check each state's uncertain tokens one by one. In this PR, it will collect the global uncertain tokens, and check only for once. It can lead to efficiency improvement especially there are a lot of states.


Seven-Streams:
Closed since no outstanding efficiency improvement is shown.

PR #448, opened 2025-10-02
Add structure tag parsing

Rocketknight1:
Extremely draft PR for now, mostly for discussion! This will definitely be changed and added to a lot before it's merged.

The idea here is that we extend the base structure tag classes with an optional `ParserTag` field. This lets us annotate blocks of text, JSON, or regex groups with ID tags. The parser schema can then use something like simple `jmespath` expressions to collect these, returning parsed output.

TODO:

- [X] Add parsing metadata to structure tag classes (drafted, might need to modify it or add it to a couple of others)
- [x] Figure out if we want a separate output structure that loads values from the parser metadata, or if the parser metadata directly specifies output paths (direct paths for now, maybe with an optional JMESPath combinator at the end?)
- [ ] Write parser logic (delegating that one to Codex :sweat_smile:)
- [ ] Tests tests tests

### DIFF
--- a/python/xgrammar/structural_tag.py
+++ b/python/xgrammar/structural_tag.py
@@ -1,7 +1,10 @@
 """Defines all structural tag formats."""
 
 import json
-from typing import Any, Dict, List, Literal, Type, Union, Optional
+import re
+from dataclasses import dataclass
+from json import JSONDecoder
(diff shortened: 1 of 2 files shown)


Rocketknight1:
Got codex to draft a parser, will start testing and iterating on it

PR #451, opened 2025-10-06
[Fix] Fix 'URIToRule'.

Seven-Streams:
This PR fixes a bug in `URIToRule`. In the original implementation, there will be a bunch of uncessary copy of `part`, and this PR uses a pointer to address the problem.


Ubospica (approved):
Use ref

PR #453, opened 2025-10-10
[Feature] Support compile_structural_tag and from_structural_tag with a Format as the parameter.

Seven-Streams:
This PR supports `compile_structural_tag` and `from_structural_tag` with a Format as the parameter. For example, in the main branch, if we want to use `from_structrual_tag` with a `StructuralTag`, we must:
```
stag_format = ...
stag = xgr.StructuralTag(format=stag_format)
grammar = xgr.Grammar.from_structural_tag(stag)
...
```

Now, we can:
```
stag_format = ...
grammar = xgr.Grammar.from_structural_tag(stag_format)
```
And it should be more convinient.

### DIFF
--- a/python/xgrammar/grammar.py
+++ b/python/xgrammar/grammar.py
@@ -1,7 +1,7 @@
 """This module provides classes representing grammars."""
 
 import json
-from typing import Any, Dict, List, Optional, Tuple, Type, Union, get_args, overload
+from typing import Any, Dict, List, Optional, Tuple, Type, Union, overload
 
 from pydantic import BaseModel
 from typing_extensions import deprecated


tests/python/test_structural_tag_converter.py
```diff
@@ -1955,8 +1986,16 @@ def test_from_structural_tag_with_structural_tag_instance(
...
+    check_stag_with_instance_in_stag_style(stag, instance, is_accepted)
+
+
+@pytest.mark.parametrize(
+    "stag_format, instance, is_accepted", basic_structural_tags_instance_is_accepted
+)
+def test_from_structural_tag_with_format_instance(
+    stag_format: xgr.structural_tag.Format, instance: str, is_accepted: bool
+):
+    check_stag_with_instance_in_format_style(stag_format, instance, is_accepted)
```
Ubospica, line 1999: Update with check_stag_with_instance

Ubospica:
Close this because we need to send StructuralTag instead of Format to compiler

PR #463, opened 2025-11-02
[CI] Add a ci to run unit tests every day.

Seven-Streams:
CI will fail caused to the request limits of HuggingFace. Although we have cached the data we need, GitHub's cache will expire after 7 days. This PR adds a CI to run the unit tests every day. With the CI, we can:
- Flush the expiration date of the cache.
- Find the potential dependency problem as soon as possible.


Ubospica (approved):
We need a daily speed test as well.

PR #475, opened 2025-11-13
[Misc] Async copy to device in apply_token_bitmask_inplace_triton

southfreebird:
Hello team,
Thank you for your incredible work!

I have noticed that you are using synchronous copying instead of asynchronous one in the Triton kernel code. This can lead to extra synchronisation, which is important for inference frameworks like VLLM. 

In this PR, I suggest switching to asynchronous copying. Please correct me if I am mistaken or if you believe it wouldn't be suitable.
Also, let me know if you need any additional tests or further assistance.


Ubospica (approved):
LGTM! This totally makes sense to me.

Could you update the format with 
```
pre-commit install
pre-commit run -a
ruff check
```

Ubospica:
Thanks! @southfreebird

PR #482, opened 2025-11-27
[Web] Support structural tag in web-xgrammar package, upgrade version to 0.1.27

akaashrp:
1. Support structural tag in web-xgrammar package
2. Upgrade version to 0.1.27
3. Upgrade devDependencies and migrate to new eslint config format
4. Fix testing


cmake/config.cmake
```diff
@@ -1,5 +1,7 @@
 set(CMAKE_BUILD_TYPE RelWithDebInfo)
-set(XGRAMMAR_BUILD_PYTHON_BINDINGS ON)
+if(NOT DEFINED XGRAMMAR_BUILD_PYTHON_BINDINGS)
```
Ubospica: Can we just remove the BUILD_PYTHON_BINDINGS flag and keep XGRAMMAR_BUILD_PYTHON_BINDINGS? So we don't need the changes in config.cmake / CMakeLists.txt

Ubospica:
cc @CharlieFRuan

PR #491, opened 2025-12-16
[Core] Fix UTF-8 support in positive character classes

### LINKED ISSUE #138: Unable to match unicode chars
Hello,

It seems the current implementation is not able to match unicode chars. My goal is to build a terminal accepting any letter including accents, such `[a-zA-Zà-ÿÀ-ß0-9 ]+`, as [GBNF](https://github.com/ggerganov/llama.cpp/blob/master/grammars/README.md#non-terminals-and-terminals) seems to support it

Simple code to reproduce with `xgrammar 1.8.0`:

```python
>>> import xgrammar.testing
>>> xgrammar.testing._is_grammar_accept_string("root ::= [é]", "é", True)
/workspace/cpp/matcher_base.cc:99: Matching char: 195 "\xc3"
/workspace/cpp/matcher_base.cc:101: Previous stack: Stacks tops size: 1
Stack #0: {
id: 0, RulePosition: rule 0: root, sequence 1: ("\xe9"), element id: 0, element in string: 0, parent id: -1, ref count: 1
}
/workspace/cpp/matcher_base.cc:131: Character 195 "\xc3" Rejected
/workspace/cpp/matcher.cc:401: Matching failed after accepting 0 characters
False
```

I also tried with a pseudo-wildcard rule:

```python
>>> xgrammar.testing._is_grammar_accept_string("root ::= [\\x00-\\xff]", "é", True)
/workspace/cpp/matcher_base.cc:99: Matching char: 195 "\xc3"
/workspace/cpp/matcher_base.cc:101: Previous stack: Stacks tops size: 1
Stack #0: {
id: 0, RulePosition: rule 0: root, sequence 1: ([\0-\xff]), element id: 0, left utf8 bytes: 0, parent id: -1, ref count: 1
}
/workspace/cpp/matcher_base.cc:131: Character 195 "\xc3" Rejected
/workspace/cpp/matcher.cc:401: Matching failed after accepting 0 characters
False
```

Unless you know a workaround, I would be happy to help solving this

Thanks a lot for your help

vasia123:
Fixes #138: Positive character classes with non-ASCII characters (e.g., [а-я], [一-龥], [😀-😿]) were not matching correctly, while negative classes worked fine.

Root cause: In CharacterClass(), Unicode codepoints were truncated to uint8_t before adding FSM edges, losing the upper bytes.

Changes:
- Add CodepointToPackedUTF8() helper to convert Unicode codepoints to packed UTF-8 format for FSM edge construction
- Update CharacterClass() to use proper UTF-8 encoding for ranges
- Add UTF-8 support in AdvanceCharacterClass() and AdvanceCharacterClassStar() for non-FSM fallback path
- Add partial_codepoint field to ParserState for UTF-8 accumulation

Tests added for:
- Cyrillic characters (2-byte UTF-8)
- CJK characters (3-byte UTF-8)
- Emoji (4-byte UTF-8)
- Mixed UTF-8 ranges
- Single non-ASCII character classes

### DIFF
--- a/.github/workflows/build_and_release.yaml
+++ b/.github/workflows/build_and_release.yaml
@@ -32,7 +32,7 @@ jobs:
           submodules: recursive
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.22.0
+        uses: pypa/cibuildwheel@v3.3.0
         env:
           CIBW_ARCHS_MACOS: ${{ matrix.arch }}
           CIBW_ARCHS_LINUX: ${{ matrix.arch }}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ else()
     )
   else()
     set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror -Wno-pedantic -Wno-unused-parameter \
--Woverloaded-virtual -flto=auto ${CMAKE_CXX_FLAGS}"
+-Woverloaded-virtual ${CMAKE_CXX_FLAGS}"
     )
   endif()
 endif()
--- a/cpp/fsm_builder.cc
+++ b/cpp/fsm_builder.cc
@@ -781,6 +781,7 @@ class TrieFSMBuilderImpl {
   TrieFSMBuilderImpl() = default;
   std::optional<FSMWithStartEnd> Build(
       const std::vector<std::string>& patterns,
+      const std::vector<std::string>& excluded_patterns,
       std::vector<int32_t>* end_states,
       bool allow_overlap,
       bool add_back_edges
@@ -790,6 +791,7 @@ class TrieFSMBuilderImpl {
 
 std::optional<FSMWithStartEnd> TrieFSMBuilderImpl::Build(
     const std::vector<std::string>& patterns,
+    const std::vector<std::string>& excluded_patterns,
     std::vector<int32_t>* end_states,
     bool allow_overlap,
     bool add_back_edges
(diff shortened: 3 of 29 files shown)


cpp/earley_parser.h
```diff
@@ -131,7 +136,8 @@ XGRAMMAR_MEMBER_ARRAY(
     &ParserState::element_id,
     &ParserState::rule_start_pos,
     &ParserState::sub_element_id,
-    &ParserState::repeat_count
+    &ParserState::repeat_count,
+    &ParserState::partial_codepoint
```
Seven-Streams, line 140: Also need to update `kXGrammarSerializeVersion` in `cpp/support/json_serializer.h` from `v7` to `v8`. Besides, some tests should be added to `test_serialization.py`.

tests/python/test_grammar_matcher_ebnf.py
```diff
@@ -500,6 +500,144 @@ def test_character_class_star_utf8():
...
+    # Multiple emojis
+    assert _is_grammar_accept_string(grammar, "😀😃😄")
+
+    # Should reject non-matching characters
+    assert not _is_grammar_accept_string(grammar, "hello")  # ASCII
+    assert not _is_grammar_accept_string(grammar, "🌍")  # Different emoji range
+    assert not _is_grammar_accept_string(grammar, "")  # empty
+
+
+def test_positive_utf8_character_class_mixed_ranges():
```
Seven-Streams, line 600: Please also add a test for the ebnf grammar:

```
"root ::= [a-zа-я一-龥]{0, 2048}"
```

Besides, could you add a test to compare the current masks with the previous ones?

Seven-Streams:
I'll check it.

cpp/earley_parser.cc
```diff
@@ -596,44 +596,95 @@ void EarleyParser::AdvanceCharacterClass(
...
+  // The state is matching a UTF8 character (continuation bytes).
   if (state.sub_element_id > 0) {
     if ((ch & 0xC0) == 0x80) {
       auto new_state = state;
       new_state.sub_element_id--;
+      // Accumulate the codepoint from continuation byte
+      new_state.partial_codepoint = (new_state.partial_codepoint << 6) | (ch & 0x3F);
+
       // Check if the UTF8 character is completed.
       if (new_state.sub_element_id == 0) {
```
Seven-Streams, line 610: I think we also need to consider the case that `new_state.sub_element_id != 0`, because it is possible that some characters should be rejected after accepting the first utf-8 character, which needs a similar logic like L611 - L641.

cpp/earley_parser.cc
```diff
@@ -660,40 +711,93 @@ void EarleyParser::AdvanceCharacterClassStar(
...
```
Seven-Streams, line 745: The same suggestion as L610's.

avanturist888:
@Seven-Streams 

Thank you for the review! I've addressed the feedback:

1. Updated `kXGrammarSerializeVersion` from v7 to v8
2. Updated version references in `test_serialization.py`
3. Added test `test_positive_utf8_character_class_with_quantifier` for `[a-zа-я一-龥]{0,2048}`
4. Added early rejection check for partial UTF-8 sequences in both `AdvanceCharacterClass` and `AdvanceCharacterClassStar` when `sub_element_id != 0`

Regarding the mask comparison test - could you clarify what you'd like compared? The existing UTF-8 tests verify correct accept/reject behavior. Did you mean a test similar to `test_fill_next_token_bitmask` that checks specific rejected token counts?

Seven-Streams:
@vasia123 Exactly. I'm wondering if you can add a test for Unicode characters similar to `test_fill_next_token_bitmask`, to check the rejected token counts. But considering that you contribute a new feature, maybe we don't need to add one. @Ubospica cc.

avanturist888:
@Seven-Streams Done! I've added test_fill_next_token_bitmask_unicode_char_class that tests rejected token counts for a grammar with mixed UTF-8 character classes ([a-zа-я一-龥]+). It verifies the bitmask generation for input containing ASCII, Cyrillic (2-byte), and CJK (3-byte) characters.

Ubospica:
@vasia123 Thanks!

PR #494, opened 2025-12-18
[Feature]  Add env to control if torch compile

xyDong0223:
# Add env to control if torch compile
When a function call is enabled in VLLM, the V1 engine defaults to enabling torch compilation. Then, upon reaching Xgrammar, it triggers the backend torch compilation, causing a crash. Unfortunately, currently some backends do not support direct integration and optimization with the PyTorch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,9 +56,16 @@ else()
     set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror -Wno-pedantic -Wno-unused-parameter \
 -Woverloaded-virtual ${CMAKE_CXX_FLAGS}"
     )
+  elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "ppc64le")
+    message(
+      STATUS "PowerPC 64-bit LE target detected, downgrading -Wfree-nonheap-object to warning."
+    )
+    set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror -Wno-pedantic -Wno-unused-parameter \
+-Woverloaded-virtual -Wno-error=free-nonheap-object -flto=auto ${CMAKE_CXX_FLAGS}"
+    )
   else()
     set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror -Wno-pedantic -Wno-unused-parameter \
--Woverloaded-virtual ${CMAKE_CXX_FLAGS}"
+-Woverloaded-virtual -Wno-error=free-nonheap-object -flto=auto ${CMAKE_CXX_FLAGS}"
     )
   endif()
 endif()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,9 +142,7 @@ skip = [
   "cp36-*",
   "cp37-*",
   "cp38-*",
-  "pp*",
   "*musllinux*",
-  "cp313-manylinux_aarch64",
   "cp313-macosx_x86_64",
 ] # pypy doesn't play nice with pybind11
 build-frontend = "build[uv]"


Seven-Streams:
Thanks for your contribution! btw, could you format the code with:
```
pre-commit install
pre-commit run -a
ruff check
```

Then I can re-run the ci.

xyDong0223:
Sure, no problem! I’ll run:

- pre-commit install
- pre-commit run -a
- ruff check

and push the updated commit shortly. Thanks!

xyDong0223:
I format the code, Thank you so much！！

Ubospica (commented):
@xyDong0223 Thanks for the contribution! Adding a torch native kernel is definitely helpful for supporting non-cuda backends.

Regarding the API design, to avoid introducing too many env variables, shall we add a `backend` parameter to the apply_token_bitmask_inplace function? The API would be like:

```
def apply_token_bitmask_inplace(
    logits: torch.Tensor,
    bitmask: torch.Tensor,
    *,
    vocab_size: Optional[int] = None,
    indices: Optional[List[int]] = None,
    backend: Literal["auto", "cpu", "cuda", "triton", "torch_compile", "torch_native"] = "auto",
) -> None:
    ...
```

where "auto" corresponds to the current dispatching logic, and each specific backend picks specific kernels.

On the vllm side, we can also call a specific (i.e. torch_native) kernel with this API to avoid crashes.

How does that sound? If you are busy, I am happy to implement it on top of your PR.

xyDong0223:
Its sound so cool! Sure, you could help me to fix my pr, since the cool idea from your mind!

Seven-Streams:
@xyDong0223 I've opened a pull request on your repo, and you can check it.

xyDong0223:
Thanks a lot for the help! 🙌
Really appreciate you implementing this on top of my PR.

python/xgrammar/matcher.py
```diff
@@ -128,30 +129,58 @@ def apply_token_bitmask_inplace(
     indices : Optional[List[int]], default: None
         A list of indices to specify which logits in the batch to apply the bitmask to. Should be
         unique. If None, apply the bitmask to all logits in the batch.
+
+    backend : Literal["auto", "cpu", "cuda", "triton", "torch_compile", "torch_native"], default: "auto"
+        The backend where the token bitmask should be applied inplace. If the value is "auto", then it will
+        choose the backend according to the type of the logits.
```
Ubospica, line 135:
```suggestion
        choose the backend according to the type of the logits.
        
        * CPU: CPU implementation
        * Triton: Default CUDA implementation
        * Torch Compile: If the hardware is not the above, use this. This supports multiple backends, including CPU/CUDA/TPU/ROCm
        * CUDA: CUDA native implementation, helpful for C++-based inference engines
        * Torch Native: Support more backends such as Kunlun. Its performance may be worse than torch compile.
```

Ubospica (approved):
Thanks @xyDong0223 @Seven-Streams !

python/xgrammar/matcher.py
```diff
@@ -128,30 +129,64 @@ def apply_token_bitmask_inplace(
...
+
```
Seven-Streams:
``` suggestion

```

python/xgrammar/matcher.py
```diff
@@ -128,30 +129,63 @@ def apply_token_bitmask_inplace(
     indices : Optional[List[int]], default: None
         A list of indices to specify which logits in the batch to apply the bitmask to. Should be
         unique. If None, apply the bitmask to all logits in the batch.
+
+    backend : Literal["auto", "cpu", "cuda", "triton", "torch_compile", "torch_native"], default: "auto"
+        The backend where the token bitmask should be applied inplace. If the value is "auto", then it will
+        choose the backend according to the type of the logits.
+        * CPU: CPU implementation
```
Seven-Streams, line 137:
``` suggestion

        * CPU: CPU implementation
```

PR #502, opened 2025-12-25
feat: provide excluded strs for any_text and tagdispatch.

Seven-Streams:
Commonly, we need some words not to appear in `any_text` or `TagDispatch`. This PR provides a new field `excluded_str` to handle the problem. For example:

```json
{
  "type": "any_text",
  "excluded_str": ["<\tag>"]
}
```

and 

```python
"""
root ::= TagDispatch(
...
excluded_str=("<\tag>")
)
"""
```

can prevent the llm outputing the `<\tag>` during the `any_text` and `TagDispatch`. The new method to build fsm(with `add_back_edge=true`) is:
1. Build the trie for patterns and excluded patterns, and collect all the end states of the excluded patterns.
2. Build back edges.
3. Remove all the edges to the end states of the excluded patterns.

### DIFF
--- a/cpp/fsm_builder.cc
+++ b/cpp/fsm_builder.cc
@@ -832,11 +832,6 @@ std::optional<FSMWithStartEnd> TrieFSMBuilderImpl::Build(
     }
   }
 
-  // TODO(Linzhang Li): Handle excluded patterns.
-  // TODO 1: Also build trie with excluded patterns.
-  // TODO 2: Build back edges as usual.
-  // TODO 3: Remove the edges to excluded end states.
-
   std::unordered_set<int32_t> dead_state_set;
 
   if (add_back_edges) {


python/xgrammar/structural_tag.py
```diff
@@ -71,6 +71,9 @@ class AnyTextFormat(BaseModel):
     type: Literal["any_text"] = "any_text"
     """The type of the format."""
 
+    excluded_str: List[str] = []
```
Ubospica:
```suggestion
    excludes: List[str] = []
```

tests/python/test_grammar_parser_macro.py
```diff
@@ -75,7 +77,8 @@ def test_lookahead_assertion_analyzer_tag_dispatch():
   ("tag5", rule5),
   stop_eos=true,
   stop_str=(),
-  loop_after_dispatch=true
+  loop_after_dispatch=true,
+  excluded_str=()
```
Ubospica:
```suggestion
  excludes=()
```

Ubospica (approved):
LGTM!

PR #503, opened 2025-12-25
[Feature] Allow empty separator string for `tags_with_separator`

ricohasgit:


cpp/structural_tag.cc
```diff
@@ -1034,13 +1031,17 @@ Result<int, ISTError> StructuralTagGrammarConverter::VisitSub(const TagsWithSepa
   auto end_str_sequence_id = end_str_expr_id == -1
                                  ? grammar_builder_.AddEmptyStr()
                                  : grammar_builder_.AddSequence({end_str_expr_id});
+
+  // Build the sequence for the recursive case, handling empty separator
+  std::vector<int> sub_sequence_elements;
```
Seven-Streams: I'm not sure if we need any special handling here. Considering that we support adding an empty bytestring in grammar_builder.h:L98. Besides, we will optimize the empty bytestring in grammar_functor.cc:L238, which will be called since a grammar is generated.

ricohasgit: Should be addressed in latest commit. Can you take another look?

Seven-Streams:
@ricohasgit Sorry for my mistake. The empty bytestring will be removed during compilation by `ByteStringFuser`(grammar_functor.h:L301). Can you modify the test?

ricohasgit:
> @ricohasgit Sorry for my mistake. The empty bytestring will be removed during compilation by `ByteStringFuser`(grammar_functor.h:L301). Can you modify the test?

Revert is up!

Ubospica:
@ricohasgit Thank you so much!

ricohasgit:
@Ubospica No problem!

PR #504, opened 2025-12-26
[Feature] Support multiple end tokens for `tag` format

ricohasgit:


cpp/structural_tag.cc
```diff
@@ -861,21 +906,37 @@ Result<int, ISTError> StructuralTagGrammarConverter::VisitSub(const TriggeredTag
...
+        );
+      } else {
+        // Handle multiple end strings for each tag
+        for (const auto& end_str : tag.end) {
+          auto end_expr_id = grammar_builder_.AddByteString(end_str);
+          choice_elements.push_back(
+              grammar_builder_.AddSequence({begin_expr_id, rule_ref_expr_id, end_expr_id})
+          );
+        }
+      }
```
Ubospica, line 983: For an empty end tag, shall we just skip adding the empty end string?

For multiple end tags, shall we 1) create another rule which equals Choice(Sequence(end1), Sequence(end2), ...), and use this rule in the choice elements (i.e. Sequence(begin_expr_id, rule_ref_expr_id, RuleRef(new_rule_id)) )?

cpp/structural_tag.cc
```diff
@@ -861,21 +906,37 @@ Result<int, ISTError> StructuralTagGrammarConverter::VisitSub(const TriggeredTag
...
       auto sub_rule_id = grammar_builder_.AddRuleWithHint("triggered_tags_sub", choice_expr_id);
       auto ref_sub_rule_expr_id = grammar_builder_.AddRuleRef(sub_rule_id);
-      auto end_str_expr_id = grammar_builder_.AddByteString(format.detected_end_str_.value());
-      auto sequence_expr_id = grammar_builder_.AddSequence({ref_sub_rule_expr_id, end_str_expr_id});
-      choice_expr_id = grammar_builder_.AddChoices({sequence_expr_id});
+      // Create choices for each detected end string
+      std::vector<int> end_choices;
+      for (const auto& end_str : format.detected_end_strs_) {
+        auto end_str_expr_id = grammar_builder_.AddByteString(end_str);
+        end_choices.push_back(grammar_builder_.AddSequence({ref_sub_rule_expr_id, end_str_expr_id})
```
Ubospica: ditto. Shall we first create a rule with a choice of end strings?

cpp/structural_tag.cc
```diff
@@ -896,11 +957,22 @@ Result<int, ISTError> StructuralTagGrammarConverter::VisitSub(const TriggeredTag
...
+        );
+      } else {
+        // Handle multiple end strings for each tag
+        for (const auto& end_str : tag.end) {
+          int end_expr_id = grammar_builder_.AddByteString(end_str);
+          choice_elements.push_back(
+              grammar_builder_.AddSequence({begin_expr_id, rule_ref_expr_id, end_expr_id})
+          );
+        }
+      }
```
Ubospica, line 1060: ditto

cpp/structural_tag.cc
```diff
@@ -927,11 +1006,22 @@ Result<int, ISTError> StructuralTagGrammarConverter::VisitSub(const TriggeredTag
...
+        );
+      } else {
+        // Handle multiple end strings for each tag
+        for (const auto& end_str : tag.end) {
+          auto end_expr_id = grammar_builder_.AddByteString(end_str);
+          first_choice_elements.push_back(
+              grammar_builder_.AddSequence({begin_expr_id, rule_ref_expr_id, end_expr_id})
+          );
+        }
+      }
```
Ubospica, line 1121: ditto

Ubospica (commented):
Thanks, LGTM! We can merge this after resolving the comments

cpp/structural_tag.cc
ricohasgit: Adding a new rule makes sense. I also changed the `tag.end.empty` case to skip adding the empty `end_expr_id` and fixed one of the tests for that as well.

Ubospica: SG. Thanks!

PR #477, opened 2025-11-21
[Feature] Support structural tag templates.

Seven-Streams:
This PR supports the structural tag templates, provides a method to generate a grammar from a structural tag template and the given values, and provides built-in structural tag templates for several models' tool-calling formats.

PR #512, opened 2026-01-04
[feature] Format level re-use for structural tags and sequence flattening

ricohasgit:
Tested with cross-grammar caching change and improvements in compilation times are complementary for long chain workflows. Could be potential overlap with #477 .


Ubospica:
This feature has be implemented in #553. Thanks again for your contribution!

Issue #500, opened 2025-12-24
[Feature] Support for 3.14 & 3.14t


lysnikolaou:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The latest version does not have this feature.
- [x] 3. I have filled all the required information below.
- [x] 4. The feature request is in English.

### requested feature

With 3.14 released, xgrammar can be easily upgraded to support it. Additionally, 3.14 marks the free-threaded build of Python as non-experimental anymore. Some initial testing shows that that can be relatively straightforwardly supported as well.

One issue is that `mlx-lm` & `mlx` crash under the free-threaded build. `mlx` can be easily made into an optional dependency with a few changes:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index f8a11ec..b962f0a 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,6 @@ dependencies = [
   "torch>=1.10.0",
   "transformers>=4.38.0",
   "triton; platform_system == 'Linux' and platform_machine == 'x86_64'",
-  "mlx-lm; platform_system == 'Darwin' and platform_machine == 'arm64'",
   "numpy",
   "typing-extensions>=4.9.0",
 ]
@@ -39,6 +38,7 @@ test = [
   # https://github.com/huggingface/transformers/issues/36906
   "transformers<4.50.0; platform_system == 'Darwin'",
 ]
+metal = ["mlx-lm; platform_system == 'Darwin' and platform_machine == 'arm64'"]
 
 [build-system]
 requires = ["scikit-build-core>=0.10.0", "nanobind==2.5.0"]
diff --git a/tests/python/test_token_bitmask_operations.py b/tests/python/test_token_bitmask_operations.py
index 95618e7..fa03106 100644
--- a/tests/python/test_token_bitmask_operations.py
+++ b/tests/python/test_token_bitmask_operations.py
@@ -7,6 +7,13 @@ from typing import Callable, List, Optional, Tuple
 import pytest
 import torch
 
+try:
+    import mlx_lm  # noqa: F401
+
+    MLX_AVAILABLE = True
+except ModuleNotFoundError:
+    MLX_AVAILABLE = False
+
 import xgrammar as xgr
 from xgrammar.testing import (
     _get_masked_tokens_from_bitmask,
@@ -16,7 +23,7 @@ from xgrammar.testing import (
 )
 
 _is_cuda_available = torch.cuda.is_available()
[...]

PR #514, opened 2026-01-05
feat: opt into free-threading in nanobind extension

lysnikolaou:
Related to #500.


rgommers:
LTO can still be enabled in the same way as before, it's orthogonal to free-threading.

Ubospica:
@lysnikolaou Thank you so much!

PR #515, opened 2026-01-07
feat: Allow any_text and triggered_tags format to terminate at any time when excludes are specified

Seven-Streams:
Previously, the structural tag analyzer will find the terminating string for any_text and triggered_tags formats. The terminating string is determined by the "end" part of the tag format wrapping the any_text or triggered_tags format. The terminating string will determine when to exit the any_text tag. If it's not generated, the any_text tag will never exit. If the terminating string is not determined, this any_text or triggered_tags will be called "unlimited". Such tag can only exist at the end of the whole structural tag and indicates can terminate at any place.

There are currently increasing needs for exclude strings for the free part of any_text or triggered_tags. There is a common use pattern:
```
Sequence(AnyTextFormat(excludes="<next_trigger>"), TagFormat(begin="<next_trigger>", content=..., end=...)
```

This essentially uses the <next_trigger> to separate any_text and the next tag. 

To support this, this PR enhances the structural tag analyzer to allow any_text and triggered_tags which:
1. Cannot detect terminating string, and
2. Has exclude strings

To terminate at any time.

### DIFF
--- a/cpp/structural_tag.cc
+++ b/cpp/structural_tag.cc
@@ -592,7 +592,6 @@ bool StructuralTagAnalyzer::IsExcluded(const Format& format) {
         } else if constexpr (std::is_same_v<T, TriggeredTagsFormat>) {
           const auto& triggered_tags_format = std::get<TriggeredTagsFormat>(format);
           return !triggered_tags_format.excludes.empty();
-          return true;
         } else {
           return false;
         }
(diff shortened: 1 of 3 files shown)


Seven-Streams:
\gemini review

Ubospica (approved):
LGTM!

PR #528, opened 2026-02-03
refactor: refactor the structure of structural_tag for better extensibility.

Seven-Streams:
This PR removes `QwenXmlParameterFormat`, and moves its logic to `JSONSchemaFormat` with `style`. The original API(`"type": "qwen_xml_parameter"`) is still usable for backward compatibility. This can improve the extensibility of the extension formats of JSONSchemaFormats. 

The original `QwenXmlParameterFormat` is not recommended, and now `JSONSchemaFormat` with `"style": "qwen_xml"` is more recommended.

PR #527, opened 2026-02-02
refactor: refactor the structure of json_schema_converter for better extensibility.

Seven-Streams:
This PR refactors the structure of the json_schema_converter to support more flexible grammars generated by the JSONschemas.

Co-authored-by: Yixin Dong <ubospica@gmail.com>

Issue #518, opened 2026-01-12
[Feature] Support Minimax XML format, similar to `qwen_xml_parameter`.


chaunceyjiang:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The latest version does not have this feature.
- [x] 3. I have filled all the required information below.
- [x] 4. The feature request is in English.

### requested feature

Support Minimax XML format, similar to `qwen_xml_parameter`.

```
<minimax:tool_call>
  <invoke name="xx">
    <parameter name="xx">xx</parameter>
....
  </invoke>
....
</minimax:tool_call>

```
example:
```
<minimax:tool_call>
  <invoke name="get_current_weather">
    <parameter name="city">Dallas</parameter>
    <parameter name="state">TX</parameter>
    <parameter name="unit">fahrenheit</parameter>
  </invoke>
  <invoke name="get_weather_forecast">
    <parameter name="city">Dallas</parameter>
    <parameter name="state">TX</parameter>
    <parameter name="unit">fahrenheit</parameter>
  </invoke>
</minimax:tool_call>

```

### Motivation

Support Minimax XML format, similar to `qwen_xml_parameter`.

mgoin:
cc @Ubospica is this a known issue for structural tag?

Seven-Streams:
I do not think we have supported it. I think I can support it after #477 is merged.

Ubospica:
@chaunceyjiang Thanks for mentioning this! We will provide support soon.

Ubospica:
Check out #529

PR #529, opened 2026-02-03
feat: support MinimaxXmlFormat and DeepSeekXmlFormat.

Seven-Streams:
As requested in #518, this PR is based on #527 and #528 and should be merged after they are merged.

This PR provides a new `style`: `minimax_xml` / `deepseek_xml` / `qwen_xml` / `json` for `JSONSchemaFormat`. Now, with the following code:

```json
{
    "type": "json_schema",
    "json_schema": {
        "type": "object",
        "required": ["name", "age"],
    },
   "style": "minimax_xml",
}
```

can be used to constrain the LLM to output in Minimax's tool-calling format. In the same way, we can use it to constrain the output in the DeepSeek-v3.2's tool-calling format.

### DIFF
--- a/cpp/json_schema_converter_ext.cc
+++ b/cpp/json_schema_converter_ext.cc
@@ -17,28 +17,15 @@ const std::string XMLToolCallingConverter::kXMLString = "xml_string";
 const std::string XMLToolCallingConverter::kXMLAny = "xml_any";
 const std::string XMLToolCallingConverter::kXMLObject = "xml_object";
 const std::string XMLToolCallingConverter::kXMLVariableName = "xml_variable_name";
-const std::unordered_map<JSONFormat, std::unordered_map<std::string, std::string>>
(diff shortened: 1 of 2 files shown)


cpp/json_schema_converter_ext.cc
```diff
@@ -0,0 +1,220 @@
...
+const std::string XMLToolCallingConverter::kXMLString = "xml_string";
+const std::string XMLToolCallingConverter::kXMLAny = "xml_any";
+const std::unordered_map<JSONFormat, std::unordered_map<std::string, std::string>>
+    XMLToolCallingConverter::kKeyWrapperMap = {
+        {JSONFormat::kQwenXML,
+         {
+             {"key_wrapper_prefix", "<parameter="},
+             {"key_wrapper_suffix", ">"},
+             {"parameter_suffix", "</parameter>"},
+         }},
```
Ubospica: struct XMLFormatStrings {string key_wrapper_prefix xxx; ...};

Issue #544, opened 2026-02-28
[Feature] Allow using a RegexFormat like object for the TagFormat.begin field.


eraser00:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The latest version does not have this feature.
- [x] 3. I have filled all the required information below.
- [x] 4. The feature request is in English.

### requested feature

Allow using a `RegexFormat` like object for the `TagFormat.begin` field.

Here is a use case describing this feature:
```
(
        xgammar.structural_tag.TagFormat(
            begin=xgammar.structural_tag.RegexFormat(pattern='<b id="\\d+">'),
            content=xgammar.structural_tag.AnyTextFormat(),
            end="</b>",
        ),
       # The grammar should match the content below:
        '<b id="1">text</b>',
       # Matching result
        True,
)
```

### Motivation

The kimi-k2 and kimi-k2.5 series LLMs use a unique tool_calls ID format like `functions.bash:0`.

They also include the tool_calls ID in the chat template (see https://huggingface.co/moonshotai/Kimi-K2.5/blob/main/chat_template.jinja), which means the number in the tool_calls ID strongly influences the model's tool call output.

When the LLM outputs parallel tool calls in a single response, the number in the tool_calls ID is hard to correctly constrain via xgrammar now.

Although we could use multiple TagFormat objects to handle this, like:
```
[TagFormat(begin=f"{i}", ...) for i in range(MAX_PARALLEL_TOOL_CALLS)]
```
The current solution is too cumbersome for this issue. 

Therefore, we request a feature to allow using a `RegexFormat` object in `TagFormat.begin`.

eraser00:
@Seven-Streams 
Could you please help review this PR: https://github.com/mlc-ai/xgrammar/pull/545?
We noticed that `cpp/structural_tag.cc` is currently drafting a major version update, and we would appreciate it if you could share your thoughts on this feature with us in advance.

Seven-Streams:
I'm reviewing it

eraser00:
Here's the current solution: https://github.com/mlc-ai/xgrammar/pull/562

PR #530, opened 2026-02-04
feat: provides builtin structural tag formats for several models.

Seven-Streams:
This PR supports several structural tag formats for different models(similar to #477 ), such as llama3.2, qwen3, deepseek, and so on. With the following code,

```python
tools = [
    {"function": {"name": "get_weather", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}}}},
    {"function": {"name": "get_time", "parameters": {"type": "object", "properties": {}}}},
]

# Get the Llama-style template function and build a structural tag
fn = get_builtin_structural_tag_template_function("llama")
structural_tag = fn({"tools": tools})
grammar = Grammar.from_structural_tag(structural_tag)
```

Then the grammar can be used to constrain the output of the LLM with correct tool-calling within any-form text
For the models like qwen3, non-thinking mode and thinking mode are both supported as well.

PR #545, opened 2026-02-28
feat: support use RegexFormat as the TagFormat.begin field (#544)

eraser00:
Allow using a `RegexFormat` like object for the `TagFormat.begin` field.  #544 

Here is a use case describing this feature:
```
(
    xgammar.structural_tag.TagFormat(
        begin=xgammar.structural_tag.RegexFormat(pattern='<b id="\\d+">'),
        content=xgammar.structural_tag.AnyTextFormat(),
        end="</b>",
    ),
    # The grammar should match the content below:
    '<b id="1">text</b>',
    # Matching result
    True,
)
```

For the `TriggeredTagsFormat`, use a `RegexBegin` object to pack the trigger and the regex.
```
(
    xgammar.structural_tag.TriggeredTagsFormat(
        triggers=["<"],
        tags=[
            xgammar.structural_tag.TagFormat(
                begin=xgammar.structural_tag.RegexBegin(
                    trigger="<",
                    # The trigger text itself should not be part of the regex pattern
                    regex=xgammar.structural_tag.RegexFormat(pattern='b id="\\d+">'),
                ),
                content=xgammar.structural_tag.AnyTextFormat(),
                end="</b>",
            )
        ],
        at_least_one=False,
        stop_after_first=False,
    ),
    '<b id="1">"1"</b>,<b id="2">"2"</b>',
    True,
)
```

Better support for Kimi-k2 and Kimi-k2.5 series LLMs by correctly constraining the tool_calls ID format, thereby improving the tool call success rate.

Update `TagFormat` and `TriggeredTagsFormat` related unit test in `test_structural_tag_converter.py`. All regression test is passed in out local run.

### DIFF
--- a/cpp/structural_tag.cc
+++ b/cpp/structural_tag.cc
@@ -53,6 +53,9 @@ class StructuralTagParser {
   /*! \brief ParseTagFormat with extra check for object and the type field. */
   Result<TagFormat, ISTError> ParseTagFormat(const picojson::value& value);
   Result<TagFormat, ISTError> ParseTagFormat(const picojson::object& value);
+  Result<std::variant<std::string, RegexBegin>, ISTError> ParseTagBeginField(
+      const picojson::value& value
+  );
   Result<TriggeredTagsFormat, ISTError> ParseTriggeredTagsFormat(const picojson::object& value);
   Result<TagsWithSeparatorFormat, ISTError> ParseTagsWithSeparatorFormat(
       const picojson::object& value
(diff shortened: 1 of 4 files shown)


eraser00:
@Seven-Streams @Ubospica 

This PR is ready for review. 🙏

eraser00:
@DarkSharpness 
This PR is ready for review. 🙏

eraser00:
> I think supporting `RegexFormat` as `begin` of `TagFormat` is a good idea. However, I think maybe we do not need to introduce `RegexBegin` because it will introduce more complexity, and it seems like its function can be replaced by `RegexFormat`?

Yes, the only usage of `RegexBegin` is for trigger matching in `TriggeredTagsFormat`; 

But I think combining regex matching and the origin string prefix trigger checking would be too complicated to understand for a high-level API.

Since `TriggeredTagsFormat` is the most commonly used pattern in the inference engine (like: sglang), I chose to add a simple API helper class to handle the original trigger matching logic.

Alternatively, we could move the `trigger` field into `TagFormat`, but that would introduce concept conflict with the existing string prefix matching semantics ...

Seven-Streams:
In this case, I think maybe we can only introduce `RegexBegin` as syntactic sugar in Python end? At parsing, we can transform `RegexBegin` into `RegexFormat`. Besides, in XGrammar, `triggers` means when the LLM outputs one of them, the llm **must** output conforming to the given structure, while we do not force the LLM output the trigger. Thus, I think `trigger` may be a little improper. @Ubospica

eraser00:
> In this case, I think maybe we can only introduce `RegexBegin` as syntactic sugar in Python end? At parsing, we can transform `RegexBegin` into `RegexFormat`. Besides, in XGrammar, `triggers` means when the LLM outputs one of them, the llm **must** output conforming to the given structure, while we do not force the LLM output the trigger. Thus, I think `trigger` may be a little improper. @Ubospica

“while we do not force the LLM output the trigger ...”

I agree with you on the trigger design. 

In fact, when using `TriggeredTagsFormat`, the trigger prefix will be removed from `TagFormat.begin`, which creates a slight usage difference from the standard `TagFormat.begin`. Additionally, the generated grammar could be somewhat difficult to interpret.
 
Anyway, I believe we can maintain backward compatibility for this API design by handling regex trigger prefix matching. It will be a bit more complicated to implement—probably taking an extra 1–2 days to build the **true RegexFormat-only version** and fix the test cases.

I will try to make a new PR, while this one remains open for any review feedback.

Ubospica:
Thanks for your PR! It definitely make sense to support tool calling ids.

I feel this usecase can actually be supported with the `excludes` field of any_text and sequence, without having to introduce a regex-based tag begin.  

For this one:

```
xgammar.structural_tag.TagFormat(
        begin=xgammar.structural_tag.RegexFormat(pattern='<b id="\\d+">'),
        content=xgammar.structural_tag.AnyTextFormat(),
        end="</b>",
    ),
```

We can consider simplifying it like this:
```
SequenceFormat([
  RegexFormat(pattern='<b id="\\d+">'),
  AnyTextFormat(excludes="</b>"),
  ConstStringFormat("</b>")
])
```

About your second example can be expressed as:
```
xgammar.structural_tag.TriggeredTagsFormat(
        triggers=["<"],
        tags=[
            xgammar.structural_tag.TagFormat(
                begin="<",
                content=SequenceFormat([
                  RegexFormat(pattern='<b id="\\d+">'),
                  AnyTextFormat(excludes="</b>")
                ]),
                end="</b>",
            )
        ],
        at_least_one=False,
        stop_after_first=False,
    ),
```

Please let me know if this makes sense!

We are indeed introducing several updates to the structural tag, mainly for 1) introducing token-level dispatch complement to the text-based; 2) introducing more formats to make structural tag easier to write. So feel free to let me know if you have any further thoughts on enhancing the structural tag.

eraser00:
@Ubospica 

We also noticed that `SequenceFormat` can be used to implement similar ID counting functionality, and the two approaches are functionally equivalent. 

However, considering the patterns used by popular LLM inference engines with XGrammar, the content portion is typically a JSON or XML schema read from different tools and model configs. 

https://github.com/vllm-project/vllm/blob/main/vllm/v1/structured_output/backend_xgrammar.py#L347
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/function_call/function_call_parser.py#L172

Adding support for `SequenceFormat` in their existing constrained decoding API would require significant changes, as different models may require different `SequenceFormat` patterns, whereas modifying only the begin field is relatively straightforward. Therefore, we raised this issue.

BTW, In my understanding, the correct implementation example of `SequenceFormat` + `TriggeredTagsFormat` should be like this ?
```
xgammar.structural_tag.TriggeredTagsFormat(
        triggers=["<"],
        tags=[
            xgammar.structural_tag.TagFormat(
                begin="<",
                content=SequenceFormat([
                  # The "<" was already accepted by the begin expr?
                  RegexFormat(pattern='b id="\\d+">'),
                  AnyTextFormat(excludes="</b>")
                ]),
                end="</b>",
            )
        ],
        at_least_one=False,
        stop_after_first=False,
    ),
```
I'm not entirely sure if my understanding is correct.

eraser00:
@Seven-Streams @Ubospica 

After reviewing the new builtin structural tag formats API at:

https://github.com/mlc-ai/xgrammar/pull/530

I believe the fastest way to support the kimi-k2 style tool call ID format is to use `SequenceFormat` in the kimi-k2 template,
and push the SGLang migration to the new API.

eraser00:
> @Seven-Streams @Ubospica
> 
> After reviewing the new builtin structural tag formats API at:
> 
> #530
> 
> I believe the fastest way to support the kimi-k2 style tool call ID format is to use `SequenceFormat` in the kimi-k2 template, and push the SGLang migration to the new API.

The current kimi-k2 template has some discrepancies with the model's chat template definition. I will try to fix the tool call ID constraint and all these discrepancy issues in a new PR.

PR #549, opened 2026-03-03
Expose IsCompleted through C++, Python, and JS bindings

ushiromiya-lion:
`is_terminated` can subtly differ from `is_complete`; this PR adds plumbing to expose `is_terminated`.


Ubospica (approved):
LGTM. Thanks!

Issue #532, opened 2026-02-06
[Feature] TVM-FFI Integration


Ubospica:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The latest version does not have this feature.
- [ ] 3. I have filled all the required information below.
- [ ] 4. The feature request is in English.

### requested feature

Currently we use nanobind to support Python bindings. TVM-FFI is a unified binding library that features:
1. Support for multiple DL frameworks (e.g. Torch, Jax, etc.)
2. Multiple language FFI support (Python, Rust, etc.)
3. Minimal overhead
4. Practical utilities (err handling, expected, etc.)

We are planning to migrate our binding system to TVM-FFI. The plan is:
1. Replace current binding system with TVM-FFI
2. Bring additional Rust support with TVM-FFI

This is expected to be completed in February.

cc @Seven-Streams @tqchen @eugenebokhan

### Motivation

This is necessary for multi-framework and multi-language support.

Seven-Streams:
The issue will be closed since #550 is merged.

PR #550, opened 2026-03-03
refactor: migrate the binding logic into `tvm_ffi`.

Seven-Streams:
As requested in #532, this PR refactors the binding logic to tvm_ffi for better extensibility and light-weight dependency.

### DIFF
--- a/python/xgrammar/kernels/apply_token_bitmask_inplace_triton.py
+++ b/python/xgrammar/kernels/apply_token_bitmask_inplace_triton.py
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import torch
 
@@ -81,7 +81,7 @@ def apply_token_bitmask_inplace_triton(
     logits: torch.Tensor,
     bitmask: torch.Tensor,
     vocab_size: Optional[int] = None,
-    indices: Optional[List[int]] = None,
+    indices: Optional[Union[List[int], torch.Tensor]] = None,
 ):
     NUM_SMS = torch.cuda.get_device_properties("cuda").multi_processor_count
     BLOCK_SIZE = 4096
@@ -106,8 +106,11 @@ def apply_token_bitmask_inplace_triton(
     num_rows = len(indices) if indices is not None else logits.shape[0] if logits.ndim == 2 else 1
 
     if indices is not None:
-        indices_cpu = torch.tensor(indices, dtype=torch.int32)
-        indices = indices_cpu.to(device=logits.device, non_blocking=True)
+        if isinstance(indices, torch.Tensor):
+            indices = indices.to(dtype=torch.int32, device=logits.device, non_blocking=True)
+        else:
+            indices_cpu = torch.tensor(indices, dtype=torch.int32)
+            indices = indices_cpu.to(device=logits.device, non_blocking=True)
 
     grid = (NUM_SMS,)
 


cpp/tvm_ffi/tvm_ffi.cc
```diff
@@ -0,0 +1,891 @@
...
+      .def_static(
+          "batch_accept_string",
+          [](ffi::Array<O> matchers_ref, ffi::Array<ffi::Any> input_str_byte_union, bool debug_print
+          ) {
+            std::vector<GrammarMatcher> storage;
+            storage.reserve(matchers_ref.size());
+            for (int64_t i = 0; i < static_cast<int64_t>(matchers_ref.size()); ++i) {
+              storage.push_back(matchers_ref[i].as<GrammarMatcherObj>()->value);
+            }
+            std::vector<std::string> strs;
```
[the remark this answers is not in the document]
Seven-Streams: not make sense. class `GrammarMatcher` only has a pointer to the impl.

DarkSharpness:
Do we still need CMake to build the project? Could we simply use `tvm_ffi.cpp.build` to build the binary and then packing them into xgrammar package?

Ubospica:
> Do we still need CMake to build the project? Could we simply use `tvm_ffi.cpp.build` to build the binary and then packing them into xgrammar package?

We still need to build the C++ library alone. I think that may not be supported well by tvm-ffi?

DarkSharpness:
> > Do we still need CMake to build the project? Could we simply use `tvm_ffi.cpp.build` to build the binary and then packing them into xgrammar package?
> 
> We still need to build the C++ library alone. I think that may not be supported well by tvm-ffi?

The tricky part is 3rdparty library dependencies. Building them with `tvm-ffi` might not be trivial.

Ubospica:
@Seven-Streams Can you show some performance results for FFI overhead before/after? It's fine if we can get similar results

python/xgrammar/compiler.py
```diff
@@ -93,7 +94,7 @@ def deserialize_json(json_str: str, tokenizer_info: TokenizerInfo) -> "CompiledG
             When the __VERSION__ field in the JSON string is not the same as the current version.
         """
         return CompiledGrammar._create_from_handle(
-            _core.CompiledGrammar.deserialize_json(json_str, tokenizer_info._handle)
+            _ffi_api.CompiledGrammar.deserialize_json(json_str, tokenizer_info._handle)
```
Ubospica: Can we keep the same _core name?

python/xgrammar/testing.py
```diff
@@ -60,7 +61,7 @@ def _json_schema_to_ebnf(
         The BNF grammar string.
     """
     schema_str = _convert_schema_to_str(schema)
-    return _core.testing._json_schema_to_ebnf(
```
Ubospica, line 63: Ditto. Can we just import all things in _core and avoid updating these lines?

python/xgrammar/testing.py
```diff
@@ -9,11 +9,12 @@
 import torch
 from pydantic import BaseModel
 
-from .base import _core
 from .compiler import CompiledGrammar, GrammarCompiler
 from .grammar import Grammar, _convert_schema_to_str
 from .matcher import GrammarMatcher, bitmask_dtype
 from .tokenizer_info import TokenizerInfo
+from .tvm_ffi_binding.testing import _ffi_api as _testing_ffi
```
Ubospica: Maybe we can still import _core

python/xgrammar/testing.py
```diff
@@ -60,7 +61,7 @@ def _json_schema_to_ebnf(
         The BNF grammar string.
     """
     schema_str = _convert_schema_to_str(schema)
-    return _core.testing._json_schema_to_ebnf(
+    return _testing_ffi._json_schema_to_ebnf(
```
Ubospica: ditto, avoid the change

python/xgrammar/config.py
```diff
@@ -19,7 +21,7 @@ def get_max_recursion_depth() -> int:
     max_recursion_depth : int
         The maximum allowed recursion depth.
     """
-    return _core.config.get_max_recursion_depth()
```
Ubospica, line 22: Ditto, we can avoid the changes and avoid adding a impl function

PR #560, opened 2026-03-11
feat: token-level structural tag API (TokenFormat, ExcludeTokenFormat, AnyTokensFormat, TokenTriggeredTagsFormat)

Ubospica:
## Summary

Add token-level structural tag formats that leverage the low-level `Token`/`ExcludeToken`/`TokenTagDispatch` grammar primitives:

- **`TokenFormat`**: match a single token by ID or string; usable in `TagFormat.begin`/`end`
- **`ExcludeTokenFormat`**: accept any single token except those in a given set (`tokens` field); auto-detects end tokens from parent `TagFormat`
- **`AnyTokensFormat`**: accept zero or more tokens except those in a given set (`exclude_tokens` field); star semantics over `ExcludeTokenFormat`; auto-detects end tokens from parent `TagFormat`
- **`TokenTriggeredTagsFormat`**: token-level tag dispatch (parallel to `TriggeredTagsFormat`), with `trigger_tokens`/`exclude_tokens` supporting both int IDs and string token names
- New `StructuralTagTokenResolver` pipeline pass to resolve string tokens to IDs via `TokenizerInfo`
- `TagFormat.begin`/`end` extended to accept `TokenFormat` variant
- Validation: `TriggeredTagsFormat` tags must use string begin; `TokenTriggeredTagsFormat` tags must use token begin; float token IDs rejected; boolean fields strictly validated
- Documentation updated with new format descriptions and token-level tool calling example

### DIFF
--- a/cpp/grammar_matcher.cc
+++ b/cpp/grammar_matcher.cc
@@ -1094,6 +1094,8 @@ void GrammarMatcher::Rollback(int num_tokens) { pimpl_->Rollback(num_tokens); }
 
 bool GrammarMatcher::IsTerminated() const { return pimpl_->IsTerminated(); }
 
+bool GrammarMatcher::IsCompleted() const { return pimpl_->IsCompleted(); }
+
 void GrammarMatcher::Reset() { pimpl_->Reset(); }
 
 int GrammarMatcher::GetMaxRollbackTokens() const { return pimpl_->GetMaxRollbackTokens(); }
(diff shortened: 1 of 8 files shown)


cpp/structural_tag.cc
```diff
@@ -703,6 +797,312 @@ Result<StarFormat, ISTError> StructuralTagParser::ParseStarFormat(const picojson
...
+    auto s = token_it->second.get<std::string>();
+    if (s.empty()) {
+      return ResultErr<ISTError>("Token string must be non-empty");
+    }
+    return ResultOk<TokenFormat>(std::variant<int32_t, std::string>(std::move(s)));
+  }
+  return ResultErr<ISTError>("TokenFormat's token must be an integer or string");
+}
+
+namespace {
```
Seven-Streams: Is it necessary to introduce another namespace?

Seven-Streams (approved):

Issue #561, opened 2026-03-11
[Feature]


eraser00:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The latest version does not have this feature.
- [x] 3. I have filled all the required information below.
- [x] 4. The feature request is in English.

### requested feature

Request better kimi-k2 series `builtin_structural_tag` support.

According to the [kimi-k2.5 chat-template](https://huggingface.co/moonshotai/Kimi-K2.5/blob/main/chat_template.jinja), we have identified two inconsistencies between the constraints defined in xgrammar's builtin_structural_tag and the model's chat template format. 

1. The format of the tool ID is `functions.{func_name}:{idx}`, from which we can parse the function name. This is critical to the correctness of the model's tool call outputs.
2. When thinking was enforce disabled, the reasoning content part should be `<think></think>`
[Like](https://huggingface.co/moonshotai/Kimi-K2.5/blob/main/chat_template.jinja#L104)

### Motivation

Use the SequenceFormat to allow regex idx
```
TagFormat(
    begin=f"<|tool_call_begin|>functions.{name}:",
    content=SequenceFormat(
        elements=[
            RegexFormat(pattern=r"\d+"),
            ConstStringFormat(value="<|tool_call_argument_begin|>"),
            JSONSchemaFormat(json_schema=parameters)
        ]
    ),
    end="<|tool_call_end|>",
)
```

Seven-Streams:
The issue will be closed since #562 is merged. Feel free to reopen it. Thanks for your contribution!

PR #562, opened 2026-03-11
Better kimi-k2 series models support in `builtin_structural_tag`

eraser00:
Fix inconsistencies between the constraints defined in xgrammar's builtin_structural_tag and the kimi-k2 series models' chat template format. (#561)

And also enable (#544) via the new builtin_structural_tag API.

Use the SequenceFormat to allow auto-incrementing numeric idx:
```
TagFormat(
    begin=f"<|tool_call_begin|>functions.{name}:",
    content=SequenceFormat(
        elements=[
            RegexFormat(pattern=r"\d+"),
            ConstStringFormat(value="<|tool_call_argument_begin|>"),
            JSONSchemaFormat(json_schema=parameters)
        ]
    ),
    end="<|tool_call_end|>",
)
```

Follow the kimi-k2-thinking and kimi-k2.5 chat templates, using pure empty reasoning content format when thinking is disabled:
```
if force_empty_reasoning:
    prefix_tag = ConstStringFormat(value="<think></think>")
else:
    prefix_tag = TagFormat(begin="<think>", content=AnyTextFormat(), end="</think>")
```

Fix all the kimi-k2 related unit test cases in `test_structural_tag_for_model.py`

All regression test in `test_structural_tag_for_model.py` is passed.

### DIFF
--- a/tests/python/test_structural_tag_for_model.py
+++ b/tests/python/test_structural_tag_for_model.py
@@ -528,6 +528,7 @@ _kimi_instances_no_tools = [
     "<think>123</think>",
     "<think>\n\n</think></think>",
     "<think></think>text",
+    "</think>123",
 ]
 
 kimi_instance_cases: List[InstanceCase] = [
@@ -694,7 +695,7 @@ any_text_1 ::= TagDispatch(
 sequence ::= ((tag any_text_1))
 root ::= ((sequence))
 """,
-        [False, False, True, False, True, True],
+        [False, False, True, False, True, False],
     ),
     # no tools, reasoning=True, force_empty_reasoning=True
     (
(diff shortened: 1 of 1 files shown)


Seven-Streams (commented):
I think Gemini's comment is right. There are 5 test instances, but 6 values in the list of expected results.

eraser00:
> I think Gemini's comment is right. There are 5 test instances, but 6 values in the list of expected results.

I add one test case in the `_kimi_instances_with_tools`, but the `_kimi_instances_no_tools` still has 5 instances.

My bad, let me fix this.

eraser00:
@Ubospica 
In order to push SGLang to fix their somewhat incorrect kimi-k2 function call parser as soon as possible, we hope this PR can be merged in the next release.

Again, many thanks for your valuable review comments!

Ubospica (approved):
Thank you so much! This support is very useful. We will release it in the next version soon.

PR #558, opened 2026-03-09
feat: support `RepeatFormat`.

Seven-Streams:
Similar to #557, this PR supports  `RepeatFormat` to enhance `StructuralTag`, and refactors the logic for repetition structures.

PR #564, opened 2026-03-14
feat: support `DispatchFormat` and `TokenDispatchFormat`.

Seven-Streams:
Similar to #558, this PR provides two `StructuralTag` to provide the semantics of `TagDispatchFormat` and `TokenTagDispatchFormat`.

### DIFF
--- a/docs/tutorials/structural_tag.md
+++ b/docs/tutorials/structural_tag.md
@@ -598,7 +598,11 @@ The format field requires a format object. We provide several basic format objec
 
     Accepts any string except the excluded string. But when any of the case strings is generated, the following format must follow the corresponding format. When the loop is false, after generating the format, we reach the end of the format. Otherwise, it will continue to allow generating any string and match the next case.
 
-    **`cases`** (required): Non-empty array of 2-element JSON arrays. Each entry is ``[trigger_string, content_format]``: a string trigger prefix, then a nested format object for the text after that trigger.
+    **`rules`** (required): Non-empty array of 2-element JSON arrays. Each entry is ``[pattern, content_format]``: the pattern is a string, followed by a format object for the content after that pattern.
+
+    **`loop`** (optional, default `true`): If true, after handling one dispatched format, it will continue to allow free-form text and match the next pattern. Otherwise, the matching of this format ends after handling the first dispatched format.
+
+    **`exclude_tokens`** (optional, default `[]`): List of strings that must not appear in the free-form text.
 
     Corresponds to an [Aho-Corasick automaton](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm) under the hood to allow efficient and accurate pattern matching in a free-formed string.
 
@@ -615,13 +619,13 @@ The format field requires a format object. We provide several basic format objec
 
 20. `token_dispatch`
 
-    Similar to `dispatch` type, while the triggers are token IDs or token strings.
+    Similar to `dispatch` type, while the patterns are token IDs or token strings.
 
-    **`cases`** (required): Non-empty array of 2-element JSON arrays. Each entry is ``[trigger_token, content_format]``: the trigger is a token ID (integer) or token string (resolved via `tokenizer_info`), followed by a format object for the content after that token.
+    **`rules`** (required): Non-empty array of 2-element JSON arrays. Each entry is ``[pattern_token, content_format]``: the pattern is a token ID (integer) or token string (resolved via `tokenizer_info`), followed by a format object for the content after that token.
 
-    **`loop`** (optional, default `true`): If true, after one dispatch the grammar allows more tokens until the next trigger.
+    **`loop`** (optional, default `true`): If true, after one dispatched format, it will continue to allow free-form text and match the next pattern. Otherwise, the matching of this format ends after handling the first dispatched format.
 
-    **`exclude_tokens`** (optional, default `[]`): Token IDs or strings to exclude before a trigger is seen.
+    **`exclude_tokens`** (optional, default `[]`): List of Tokens that must not appear in the free-form text.
 
     ```json
     {
--- a/python/xgrammar/structural_tag.py
+++ b/python/xgrammar/structural_tag.py
@@ -354,43 +354,63 @@ class RepeatFormat(BaseModel):
 
 
 class DispatchFormat(BaseModel):
-    """A format that maps directly to a TagDispatch grammar.
+    """Matches certain patterns in free-form text.
+    When certain strings are generated, the following content must follow the corresponding format.
+    Certain strings can be excluded from being generated in the free-form part.
 
-    Accepts a list of (trigger string, content format) pairs. In JSON, each pair is a two-element
-    array ``[trigger, content]``. When the output matches a trigger string, generation continues
-    with the corresponding content format. This is a lower-level alternative to
-    ``TriggeredTagsFormat`` when you want to specify exact trigger→rule mapping without tag
-    begin/end structure.
+    The user specifies a list of (pattern, formats). The LLM can generate any free-form text, but
+    when a pattern is matched in the text, the following output must follow the corresponding format.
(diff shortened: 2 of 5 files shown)


python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,62 @@ class RepeatFormat(BaseModel):
...
+
+    Accepts a list of (trigger string, content format) pairs. When the output matches a trigger
+    string, generation continues with the corresponding content format. This is a lower-level
+    alternative to ``TriggeredTagsFormat`` when you want to specify exact trigger→rule mapping
+    without tag begin/end structure.
+    """
+
+    type: Literal["tag_dispatch"] = "tag_dispatch"
+    """The type of the format."""
+    pairs: List[TagDispatchPair]
```
Ubospica:
```suggestion
    pairs: List[Tuple[str, Format]]
```

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,62 @@ class RepeatFormat(BaseModel):
...
+class TagDispatchPair(BaseModel):
+    """A (trigger string, content format) pair for TagDispatchFormat."""
+
+    trigger: str
+    """The trigger string. When this prefix is seen, generation dispatches to content."""
+    content: "Format"
+    """The format for the content after the trigger."""
+
+
+class TagDispatchFormat(BaseModel):
```
Ubospica: DispatchFormat

cases: List[tuple[str, Format]]

loop: bool
excludes: List[str]


TokenDispatchFormat
cases
loop
exclude_tokens

docs/tutorials/structural_tag.md
```diff
@@ -594,6 +594,54 @@ The format field requires a format object. We provide several basic format objec
 
     `trigger_tokens` and `exclude_tokens` elements can be integers or strings. `at_least_one` and `stop_after_first` have the same semantics as in `triggered_tags`.
 
+19. `tag_dispatch`
+
+    A format that maps directly to a **TagDispatch** grammar. It accepts a list of (trigger string, content format) pairs. When the output matches a trigger string, generation continues with the corresponding content format. This is a lower-level alternative to `triggered_tags` when you want exact trigger→rule mapping without tag begin/end structure.
```
Ubospica: Accepts any string except the excludes string. But when any of the case string is generated, the following format must follow the corresponding format. When loop is false, after generating the format, we reach the end of the format. Otherwise, it will continue allow generating any string and match the next case.

Corresponds to an Aho-corasick (wikipedia link) under the hood to allow efficient and accurate pattern matching in free-formed string.

Common usage: Put the end string inside the exclude, and put another end string after it. E.g.
```json
format: sequence
sequence: [
DispatchFormat(cases=..., excludes=["abc"])
ConstStringFormat("abc")
]
```

This means when the LLM generates abc, the matching of the first dispatchformat ends and gets into the following conststring format. You can also use the "end" part of the TagFormat to achieve the same goal.

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,46 @@ class RepeatFormat(BaseModel):
...
+    Accepts a list of (trigger string, content format) pairs. In JSON, each pair is a two-element
+    array ``[trigger, content]``. When the output matches a trigger string, generation continues
+    with the corresponding content format. This is a lower-level alternative to
+    ``TriggeredTagsFormat`` when you want to specify exact trigger→rule mapping without tag
+    begin/end structure.
+    """
+
+    type: Literal["dispatch"] = "dispatch"
+    """The type of the format."""
+    cases: List[Tuple[str, "Format"]]
```
Ubospica:
```suggestion
    rules: List[Tuple[str, "Format"]]
```

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,46 @@ class RepeatFormat(BaseModel):
     """The format that is repeated."""
 
 
+class DispatchFormat(BaseModel):
+    """A format that maps directly to a TagDispatch grammar.
```
Ubospica: Semantics

Matches certain patterns in free-form text. When certain strings are generated, the following content must follow the corresponding format. Certain strings can be excluded from being generated in the free-form part.

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,46 @@ class RepeatFormat(BaseModel):
...
+    ``TriggeredTagsFormat`` when you want to specify exact trigger→rule mapping without tag
+    begin/end structure.
+    """
+
+    type: Literal["dispatch"] = "dispatch"
+    """The type of the format."""
+    cases: List[Tuple[str, "Format"]]
+    """List of ``(trigger string, content format)`` pairs."""
+    loop: bool = True
+    """If true, after handling one dispatch the grammar allows any text until the next trigger."""
```
Ubospica:
```suggestion
    """If true, after handling one dispatched format, it will continue to allow free-form text and match the next pattern. Otherwise, the matching of this format ends after handling the first dispatched format."""
```

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,46 @@ class RepeatFormat(BaseModel):
...
+    excludes: List[str] = []
+    """List of strings that must not appear before a trigger is seen."""
```
Ubospica:
```suggestion
    """List of strings that must not appear in the free-form text."""
```

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,46 @@ class RepeatFormat(BaseModel):
...
 
 
+class DispatchFormat(BaseModel):
+    """A format that maps directly to a TagDispatch grammar.
+
+    Accepts a list of (trigger string, content format) pairs. In JSON, each pair is a two-element
+    array ``[trigger, content]``. When the output matches a trigger string, generation continues
+    with the corresponding content format. This is a lower-level alternative to
+    ``TriggeredTagsFormat`` when you want to specify exact trigger→rule mapping without tag
+    begin/end structure.
```
Ubospica: The user specifies a list of (pattern, formats). The LLM can generate any free-form text, but when a pattern is matched in the text, the following output must follow the corresponding format.

The `loop` field controls whether the matching of this structure ends after accepting the first pattern and format. If False, it ends. If true, the matching continues, and this format will continue to allow free-form text and detect the next pattern to be generated.

The `excludes` field controls the strings that cannot be generated in the free-form text. It can also control the end of the format: 

SequenceFormat(DispatchFormat(..., excludes="<end>"), ConstStringFormat("<end>"))

or 

TagFormat(begin=..., content=DispatchFormat(..., excludes="<end>"), end="<end>")

ends the matching of the format when LLM generates `<end>`.

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,46 @@ class RepeatFormat(BaseModel):
...
+    """
+
+    type: Literal["dispatch"] = "dispatch"
+    """The type of the format."""
+    cases: List[Tuple[str, "Format"]]
+    """List of ``(trigger string, content format)`` pairs."""
```
Ubospica:
```suggestion
    """List of ``(pattern, content format)`` pairs."""
```

python/xgrammar/structural_tag.py
```diff
@@ -352,6 +352,46 @@ class RepeatFormat(BaseModel):
...
+    array ``[trigger, content]``. When the model generates the trigger token, generation continues
+    with the corresponding content format. This is a lower-level alternative to
+    ``TokenTriggeredTagsFormat`` when you want exact token→rule mapping without tag begin/end
+    structure.
+    """
+
+    type: Literal["token_dispatch"] = "token_dispatch"
+    """The type of the format."""
+    cases: List[Tuple[Union[int, str], "Format"]]
+    """List of ``(trigger token, content format)`` pairs. Trigger is token ID or token string."""
```
Ubospica:
```suggestion
    """List of ``(token pattern, content format)`` pairs. Trigger is token ID or token string."""
```

PR #573, opened 2026-03-26
fix: support patternProperties alongside properties in JSON schema

### LINKED ISSUE #487: [Bug] Incomplete support for json schema keywords propertyNames and patternProperties
### Checklist

### Describe the bug

The support for `propertyNames` and `patternProperties` is incomplete, which makes it impossible for vllm to use xgrammar as a backend for schemas containing those fields (see https://github.com/vllm-project/vllm/pull/30390).

### patternProperties
When using both, `properties` and `patternProperties` in the same schema, only keys that match the `patternProperties` are considered during the validation, resulting in the rejection of schema-conform json objects.

### propertyNames
When using propertyNames with generic patterns such as `"^.*$"` arbitrary llm output can be included in the structured output after the end of the json object

### patternProperties
```
import xgrammar as xgr

SCHEMA = {
    "type": "object",
    "properties": {
        "name": {"type": "string"},
        "grade": {"type": "string"},
    },
    "required": ["name", "grade"],
    "patternProperties": {
        "^grade$": {"type": "string"},
    }
}

RESPONSE_VALID = '{"name": "John Doe", "grade": "B"}'
RESPONSE_INCOMPLETE = '{"grade": "B"}'

tokenizer = AutoTokenizer.from_pretrained("gpt2")
tokenizer_info = xgr.TokenizerInfo.from_huggingface(tokenizer)
compiler = xgr.GrammarCompiler(tokenizer_info)
grammar = compiler.compile_json_schema(SCHEMA)
matcher = xgr.GrammarMatcher(grammar)

assert matcher.accept_string(RESPONSE_VALID) == True
assert matcher.accept_string(RESPONSE_INCOMPLETE) == False
```
### propertyNames
```
import xgrammar as xgr

SCHEMA = {
    "type": "object",
    "properties": {
        "name": {"type": "string"},
        "grade": {"type": "string"},
    },
    "required": ["name", "grade"],
    "propertyNames": {"pattern": "^.*$"},
}

RESPONSE = '{"name": "John Doe", "grade": "B"} and spurious llm output'

tokenizer = AutoTokenizer.from_pretrained("gpt2")
tokenizer_info = xgr.TokenizerInfo.from_huggingface(tokenizer)
compiler = xgr.GrammarCompiler(tokenizer_info)
grammar = compiler.compile_json_schema(SCHEMA)
matcher = xgr.GrammarMatcher(grammar)

assert matcher.accept_string(RESPONSE) == False
```

### Environment

- xgrammar version: 0.1.27
- serving engine: vllm
- platform: linux ubuntu

15935648a:
## Summary

- When both `properties` and `patternProperties` are defined in a JSON schema object, the grammar now correctly handles both constraints simultaneously
- Previously, `patternProperties` would cause `properties` to be ignored entirely, rejecting valid JSON objects
- Added regression tests for the fixed behavior
- Known limitation: when multiple `patternProperties` are present, value schemas are currently unioned across all patterns rather than enforced per-pattern. See the TODO comment in `test_multiple_pattern_properties` for details. This requires per-pattern key-value enforcement and is left for future work.
- Verified that the `propertyNames` trailing content issue (also mentioned in #487) is already fixed in the current version. Added a regression test `test_property_names_does_not_accept_trailing_content` to prevent future regressions.

Fixes #487

## Test plan

- [ ] `test_pattern_properties_with_properties` — basic case, both fields present
- [ ] `test_multiple_pattern_properties` — multiple patterns
- [ ] `test_pattern_properties_extra_key` — extra keys matching pattern
- [ ] `test_pattern_properties_value_type` — value type enforcement

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -1570,8 +1570,11 @@ std::string JSONSchemaConverter::GetPartialRuleForProperties(
     const std::string& rule_name,
     const std::string& additional_suffix,
     int min_properties,
-    int max_properties
+    int max_properties,
+    const std::string& additional_key_pattern
 ) {
+  const std::string& key_pat =
+      additional_key_pattern.empty() ? GetKeyPattern() : additional_key_pattern;
   if (max_properties == 0) {
     return "";
   }
@@ -1600,7 +1603,7 @@ std::string JSONSchemaConverter::GetPartialRuleForProperties(
     if (allow_additional) {
       std::string add_value_rule = CreateRule(additional, rule_name + "_" + additional_suffix);
       additional_prop_pattern =
-          FormatOtherProperty(GetKeyPattern(), add_value_rule, rule_name, additional_suffix);
+          FormatOtherProperty(key_pat, add_value_rule, rule_name, additional_suffix);
       std::string last_rule_body = "(" + mid_sep + " " + additional_prop_pattern + ")*";
       std::string last_rule_name =
           rule_name + "_part_" + std::to_string(static_cast<int>(properties.size()) - 1);
@@ -1656,7 +1659,7 @@ std::string JSONSchemaConverter::GetPartialRuleForProperties(
     if (allow_additional) {
       std::string add_value_rule = CreateRule(additional, rule_name + "_" + additional_suffix);
       additional_prop_pattern =
-          FormatOtherProperty(GetKeyPattern(), add_value_rule, rule_name, additional_suffix);
+          FormatOtherProperty(key_pat, add_value_rule, rule_name, additional_suffix);
     }
 
     // Get the range of matched properties for each rule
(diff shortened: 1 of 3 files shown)


15935648a:
Addressed Copilot's feedback: moved the two failing assertions in test_pattern_properties_with_properties and test_pattern_properties_extra_key to TODO comments. These cases expose a known limitation — when patternProperties is present, extra keys that don't match any pattern are currently accepted because GetPartialRuleForProperties falls back to GetKeyPattern() (any string) for the additional-key pattern. Fixing this would require changing the signature of GetPartialRuleForProperties to accept a pattern constraint, which is left for a follow-up.

15935648a:
Update: I went ahead and implemented the fix I described as a follow-up above.

- additional_key_pattern parameter (cpp/json_schema_converter.h:362, cpp/json_schema_converter.cc:1574): Added to GetPartialRuleForProperties so callers can restrict extra key names. Defaults to empty string (falls back to GetKeyPattern()), so all existing call sites are unaffected.

- Case 1a pattern_key (cpp/json_schema_converter.cc:1938): When no explicit additionalProperties is set, builds an EBNF key pattern from the union of all patternProperties regex patterns and passes it as additional_key_pattern. 

- Case 1c (cpp/json_schema_converter.cc:1965): New branch for properties + propertyNames coexistence. Uses CreateRule(spec.property_names, ...) as the key pattern for extra keys, so propertyNames is no longer silently ignored.

- Tests (tests/python/test_grammar_matcher_json_schema.py): Restored the two TODO-commented assertions in test_pattern_properties_with_properties and test_pattern_properties_extra_key, and added test_property_names_with_properties for Case 1c.

PR #578, opened 2026-03-27
fix: AnyTokensFormat with exclude_tokens should be treated as self-terminating

ushiromiya-lion:
`AnyTokensFormat` with non-empty `exclude_tokens` is self-terminating — content stops when an excluded token appears — but `IsExcluded` was missing this case. As a result, `token_triggered_tags` tags using `content: any_tokens(exclude=[...])` with `end: ""` were incorrectly rejected with *"When the content is unlimited, at least one end string must be non-empty"*.

Adds the missing branch to `IsExcluded`, consistent with the existing handling of `AnyTextFormat` with string excludes.


ushiromiya-lion:
@Seven-Streams @Ubospica any chance we merge this before PR rot sets in,,?

Ubospica:
@ushiromiya-lion Thank you so much!

PR #586, opened 2026-04-09
feat: support `tool_choice` for `get_builtin_structural_tag`.

Seven-Streams:
This PR enhances the function `get_builtin_structural_tag` by adding an extra parameter `tool_choice`. This PR supports 3 different `tool_choice`:

- `auto`: tool-calling will appear in any-form text, and it is optional to call tools.
- `required`: the output will be one or more tool-calling(s), without any other text.
- `forced`: the output will be exactly one tool-calling, and the tool is chosen by the user. In this case, another parameter `forced_function_name` should also be passed to `get_builtin_structural_tag`, which is the chosen tool's name(str).

With different `tool_choice` parameters, `get_builtin_structural_tag` will return the corresponding `StructuralTag`.


python/xgrammar/builtin_structural_tag.py
```diff
@@ -58,6 +62,10 @@ def get_builtin_structural_tag(
         the empty thinking content at the beginning of the response.
         Some models like Qwen3, DeepSeek-R1 and etc. prefer empty-thinking mode to disable
         reasoning mode instead of non-thinking mode.
+    tool_choice : Literal["auto", "forced", "required"]
+        How tool calls are constrained relative to the provided tools list.
```
Ubospica, line 64: required: at least one tool

forced: exactly one tool. You can pass the single expected tool in tools

python/xgrammar/builtin_structural_tag.py
```diff
@@ -111,6 +122,23 @@ def get_builtin_structural_tag_supported_models(
 ] = {}
 _structural_tag_supported_models: Dict[BuiltinSupportedModels, List[str]] = {}
 _THINK_EXCLUDE_TOKENS = ["<think>", "</think>"]
+_VALID_BUILTIN_TOOL_CHOICES: List[BuiltinStructuralTagToolChoice] = ["auto", "forced", "required"]
+FORCED_TOOL_ERROR = "The 'forced_function_name' is required when 'tool_choice' is 'forced'."
```
Ubospica:
```suggestion
_FORCED_TOOL_ERROR = "The 'forced_function_name' is required when 'tool_choice' is 'forced'."
```

python/xgrammar/builtin_structural_tag.py
```diff
@@ -111,6 +122,23 @@ def get_builtin_structural_tag_supported_models(
...
+REQUIRED_TOOLS_ERROR = (
```
Ubospica:
```suggestion
_REQUIRED_TOOLS_ERROR = (
```

python/xgrammar/builtin_structural_tag.py
```diff
@@ -28,20 +28,24 @@
     "glm47",
 ]
 
+BuiltinStructuralTagToolChoice = Literal["auto", "forced", "required"]
```
Ubospica: inline

docs/tutorials/advanced_structural_tag.md
```diff
@@ -28,8 +28,10 @@ Use it when you need to constrain the model to output in a fixed pattern such as
   Default value is `[]`.
```
Ubospica: Add a section: structural tag

1. Structural Tag API
2. Builtin STag
  1. Simple intro, usage
  2. Supported models
  1. API (link to generated doc)
4. Advanced Usages

python/xgrammar/builtin_structural_tag.py
Ubospica:
```suggestion
    model: str,
```

We can avoid hardcoding it. User code can extend the models and still use this api

Ubospica:
```suggestion
```
Let's avoid hardcoding it but put the full list in doc

Ubospica:
```suggestion
def register_builtin_structural_tag(name: str):
```
Make it public and exported in `__init__.py`; Let's just put the supported model field in docstring, like

Supported models:
* Meta-Llama-3
* Llama-3.1
* Llama-3.2
* Llama-4

Ubospica:
```suggestion
def get_llama_structural_tag(input_dict: Dict[str, Any]) -> StructuralTag:
```
These can be public function and documented but not exported

python/xgrammar/builtin_structural_tag.py
```diff
@@ -36,13 +36,14 @@ def get_builtin_structural_tag(
     tools: List[Dict[str, Any]] = [],
     builtin_tools: List[Dict[str, Any]] = [],
     force_empty_reasoning: bool = False,
+    tool_choice: Literal["auto", "required"] = "auto",
```
Ubospica:
```suggestion
    tool_choice: Union[Literal["auto", "required", "none"], 
```
required: at least once

forced: exactly once.

python/xgrammar/builtin_structural_tag.py
Ubospica: Let's remove it

PR #588, opened 2026-04-10
feat: add builtin structural tag support for Gemma 4 models

### LINKED ISSUE #587: Add builtin structural tag support for Gemma 4 models
## Summary

Gemma 4 reasoning models (gemma-4-12b-it, gemma-4-26b-a4b-it, gemma-4-31b-it, etc.) use a non-standard delimiter format for thinking/reasoning that is not currently supported by `get_builtin_structural_tag`.

Adding support would enable grammar-constrained decoding to coexist with Gemma 4's thinking mode, the same way it works today for Qwen, DeepSeek, Llama, and other supported model families.

## Gemma 4 thinking format

Gemma 4 uses channel markers instead of `<think>...</think>`:

```
<|channel>thought
...chain of thought...<channel|>Final answer here.
```

- **Open marker:** `<|channel>thought\n` (token ID 100 for `<|channel>`)
- **Close marker:** `<channel|>` (token ID 101)
- The `thought` after `<|channel>` is a role label, not a separate token

When thinking is disabled via `enable_thinking=False`, the chat template pre-closes the channel with an empty block:
```
<|channel>thought\n<channel|>
```

## Why this matters

Without structural tag support, grammar-constrained decoding and thinking are mutually exclusive for Gemma 4:

- If grammar constraints are active from the start, thinking tokens violate the grammar and generation fails or produces garbage.
- If thinking is disabled to work around this, the model loses its reasoning capability.

The structural tag would allow xgrammar to let thinking tokens pass freely, then activate grammar constraints after the `<channel|>` close marker — matching the existing pattern for other model families.

## Ecosystem context

vLLM has a similar issue: when using `--reasoning-parser gemma4` with structured output and `enable_thinking=false`, grammar constraints are silently bypassed because the reasoning end token is never detected (vllm-project/vllm#39130). A proper xgrammar structural tag for Gemma 4 would provide a cleaner solution at the grammar level.

## Suggested structural tag definition

```python
"gemma4": BuiltinStructuralTagDefinition(
    think_open="<|channel>thought\n",
    think_close="<channel|>",
    # content phase has no explicit open/close — everything after <channel|> is content
)
```

Model name patterns: `gemma-4-*-it`, `Gemma-4-*`

## Current supported models for reference

`llama`, `qwen`, `qwen_coder`, `kimi`, `deepseek_r1`, `deepseek_v3_2`, `harmony`, `minimax`, `glm47`

leuski:
## Summary

Adds builtin structural tag support for Google Gemma 4 models, enabling grammar-constrained decoding to coexist with Gemma 4's thinking mode.

Gemma 4 uses channel markers instead of `<think>`/`</think>`:
- **Thinking:** `<|channel>thought\n...<channel|>`
- **Tool calls:** `<|tool_call>call:func_name{...}<tool_call|>`

This follows the same pattern as existing model registrations (Qwen, DeepSeek, Llama, etc.).

## Changes

- **`python/xgrammar/builtin_structural_tag.py`**: Register `"gemma4"` in `BuiltinSupportedModels` and add `_get_gemma4_structural_tag()` with support for reasoning, force_empty_reasoning, and tool calls
- **`tests/python/test_structural_tag_for_model.py`**: Add instance acceptance/rejection tests covering all modes (reasoning on/off, force_empty, tools, combined), supported models listing, and no-parameter tools smoke test

## Format details

From the Gemma 4 tokenizer's `response_schema`:

```
(<|channel>thought\n...thinking...<channel|>)?    # optional thinking
(content text)?                                    # optional content  
(<|tool_call>call:name{json}<tool_call|>)*         # optional tool calls
(<turn|>)?                                         # optional turn end
```

Reference: https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4

## Test plan

- [ ] `ruff check` passes
- [ ] Instance acceptance tests for all reasoning/tool combinations
- [ ] Smoke test with no-parameter tools
- [ ] Supported models listing includes gemma4

Closes #587

### DIFF
--- a/python/xgrammar/builtin_structural_tag.py
+++ b/python/xgrammar/builtin_structural_tag.py
@@ -26,6 +26,7 @@ BuiltinSupportedModels = Literal[
     "deepseek_v3_2",
     "minimax",
     "glm47",
+    "gemma4",
 ]
 
 
@@ -111,6 +112,7 @@ _structural_tag_registry: Dict[
 ] = {}
 _structural_tag_supported_models: Dict[BuiltinSupportedModels, List[str]] = {}
 _THINK_EXCLUDE_TOKENS = ["<think>", "</think>"]
+_GEMMA4_EXCLUDE_TOKENS = ["<|channel>", "<channel|>"]
 
 
 def _validate_tool_function(tools: Any) -> None:
@@ -202,6 +204,8 @@ def _get_builtin_structural_tag_function(
           Supported Models: Deepseek-V3.1, Deepseek-R1, Deepseek-V3.2-exp and other models that follow the same style.
         - "harmony": OpenAI Harmony Response Format (gpt-oss).
           Supported Models: GPT-oss and other models that follow the same style.
+        - "gemma4": Gemma 4 style structural tag format.
+          Supported Models: Gemma-4 and other models that follow the same style.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


Seven-Streams (approved):
LGTM

PR #593, opened 2026-04-23
feat: support llama's user-defined custom tool calling format.

Seven-Streams:
This PR provides a new builtin structural tag for [llama's user-defined custom tool calling format](https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_1/#user-defined-custom-tool-calling) to enhance the builtin structural tag. This PR also fixes the issue that when `"not hf_token_required"` is passed to the pytest, the tests will still ask for a hf_token in some cases.


Ubospica:
Let's check this has been supported and we can close this

PR #595, opened 2026-04-26
fix: incorrect mask with additionalProperties:true and non-required properties

### LINKED ISSUE #208: incorrect mask with additionalProperties:true and non-required properties
With the following schema, xgrammar will accept `{"a": "wrong"}` even though it shouldn't ("a" is supposed to be an integer)

```json
{
    "type": "object",
    "properties": {
        "a": {"type": "integer"},
    },
    "additionalProperties": True,
    "required": [],
}
```

full repro

```python
import xgrammar as xgr
import torch
import numpy as np
import json

from transformers import AutoTokenizer, AutoConfig
# Get tokenizer info
model_id = "unsloth/Meta-Llama-3.1-8B-Instruct"
tokenizer = AutoTokenizer.from_pretrained(model_id)
config = AutoConfig.from_pretrained(model_id)
# This can be larger than tokenizer.vocab_size due to paddings
full_vocab_size = config.vocab_size
tokenizer_info = xgr.TokenizerInfo.from_huggingface(tokenizer, vocab_size=full_vocab_size)

compiler = xgr.GrammarCompiler(tokenizer_info, max_threads=1)

schema = {
    "type": "object",
    "properties": {
        "a": {"type": "integer"},
    },
    "additionalProperties": True,
    "required": [],
}
compiled_grammar = compiler.compile_json_schema(json.dumps(schema))
matcher = xgr.GrammarMatcher(compiled_grammar)
token_bitmask = xgr.allocate_token_bitmask(1, tokenizer_info.vocab_size)

sim_sampled_response = '{"a": "wrong"}'
sim_sampled_token_ids = tokenizer.encode(sim_sampled_response, add_special_tokens=False)

for i, sim_token_id in enumerate(sim_sampled_token_ids):
    matcher.fill_next_token_bitmask(token_bitmask)
    assert matcher.accept_token(sim_token_id)
```

ianliuy:
## What's broken?

When a JSON schema has `additionalProperties: true` with properties that are **not** in the `required` array, xgrammar accepts values with the **wrong type** for those defined properties. For example, `{"a": "wrong"}` is accepted even though `"a"` is defined as `integer`.

## Who is affected?

Anyone using `additionalProperties: true` (or `unevaluatedProperties: true`) with non-required typed properties. This includes the common pattern of optional typed fields with free-form extra data. Schemas with **all properties required** or `additionalProperties: false` are NOT affected.

## When does it trigger?

Reproducible 100% of the time with:
```json
{
    "type": "object",
    "properties": { "a": {"type": "integer"} },
    "additionalProperties": true,
    "required": []
}
```
Input `{"a": "wrong"}` is incorrectly accepted.

## Root Cause

`GetKeyPattern()` returns `basic_string`, which matches **any** JSON string key -- including defined property names like `"a"`. When the grammar adds `additional_prop_pattern` as an alternative, the parser tracks **both** paths simultaneously:

1. **Typed path**: `"a" : integer_rule` -- correctly constrains value type
2. **Catch-all path**: `basic_string : any_rule` -- accepts any value type

The token bitmask is the **union** of valid tokens across all active paths, so the catch-all's `any_rule` effectively overrides the typed constraint.

## Fix

Replace `GetKeyPattern()` with `GetKeyPatternExcluding(properties, rule_name)` at 3 call sites in `GetPartialRuleForProperties()`. The new method generates a **trie-based EBNF key pattern** that matches any JSON string key **except** defined property names.

For excluded key `"a"`, the generated pattern:
- Accepts `""`, `"b"`, `"ab"` (not the excluded key)
- Rejects `"a"` (must use typed rule instead)

## Testing

Added `test_additional_properties_type_enforcement` with 7 test cases:
1. Wrong type + empty required -> rejected (core bug fix)
2. Correct type -> accepted
3. Unknown key (truly additional) -> accepted
4. Defined + additional together -> accepted
5. Partial required, wrong type on non-required -> rejected
6. All correct types -> accepted
7. Empty object -> accepted

Fixes #208

cc @Ubospica @Seven-Streams


Ubospica (approved):
This looks good to me.

One challenge would be, in some cases, a parameter name does not end with `"`, but with a long string like `</parameter>`. Then the trie solution needs to be further extended to exclude the end string existing anywhere in the text.

I feel we can merge this since this already resolves the JSON path well. We can support other paths (e.g. qwen parameter format) later. cc @Seven-Streams it'd be great if you can double check if this will affect other paths.

PR #608, opened 2026-04-29
[Web] Use the modern `emcc -lembind` rather than `emcc --bind`

kolayne:
Replace the `--bind` option to `emmc` with `-lembind` in the build script.

According to `emcc --help`:
> "--bind"
> [link] Links against embind library.  Deprecated: Use "-lembind"
> instead.


CharlieFRuan:
@akaashrp Could you take a look when you get a chance please? Thanks!

akaashrp (approved):
LGTM! @Seven-Streams, if you could approve the workflows and take a look, that would be great. Thanks.

PR #616, opened 2026-05-01
perf: reduce structural-tag compile time at high tool count

Ubospica:
## TL;DR
On the BFCL tool-calling workload (Llama 3.1 8B, `tool_choice="required"`), `compile_structural_tag` cold time at 1000 tools drops from **21.7s → 2.1s (-90%)**.

## Bottlenecks (1000 tools, baseline)
1. **`GrammarBuilder::GetNewRuleName` is O(N) per call.** For each new rule it probes `name_hint`, `name_hint_1`, `name_hint_2`, ... from scratch against `rule_name_to_id_`. With ~11k rules sharing a small set of name hints this becomes O(N^2) overall and dominates baseline time.
2. **`LookaheadAssertionAnalyzer` is O(N) per rule.** `IsExactLookaheadAssertion` and `DetectLookaheadAssertion` each scan all rules to determine dispatch triggers, last-position references, and suffix sequences. Called for every rule, this is O(N^2) overall.
3. **`MinimizeDFA` in `Choices()` is wasted work.** State count is essentially unchanged after Hopcroft minimization (e.g. 160,002 → 160,578), but the call itself takes ~580 ms.
4. **`MergeEquivalentSuccessors` thrashes hash maps.** It rebuilds `vector<unordered_map<int, vector<FSMEdge>>>` for predecessor and successor groups every iteration. `build_maps` accounts for ~75% of its runtime, dominated by hash-map and small-vector allocations across 11k+ rules where most rules have <30 states.

### 1. `GrammarBuilder::GetNewRuleName`
Cache the next available suffix per `name_hint` so repeated calls with the same hint are amortized O(1) instead of O(N).

### 2. `LookaheadAssertionAnalyzer::BuildRuleLookaheadInfo`
Pre-compute all per-rule lookahead info (dispatch triggers, last-position references, non-last occurrence counts and suffix sequences) in a single O(N) pass over all rules. `IsExactLookaheadAssertion` and `DetectLookaheadAssertion` become O(1) lookups, eliminating the O(N^2) scanning.

### 3. Drop `MinimizeDFA` in `Choices()`
Remove the call to `MinimizeDFA()` after `Union → SimplifyEpsilon → MergeEquivalentSuccessors`. It does not reduce states on these grammars and scales super-linearly in state count.

### 4. CSR-based `MergeEquivalentSuccessors`
- Replace per-state `unordered_map<int, vector<FSMEdge>>` with two `Compact2DArray<EndpointEdge>` (incoming/outgoing edges). Two passes compute row sizes and fill rows directly; no hash map, no per-edge small allocations.
- Hoist scratch buffers (`row_sizes`, `write_positions`, `incoming_edges`, `outgoing_edges`, distinct-peer info) out of the iteration loop and reuse capacity across iterations.
- Early-exit when `NumStates() < 4`: no equivalent-successor merge is structurally possible.

### 5. `Compact2DArray` API extensions (supporting change)
- `FromRowSizes(row_sizes)` / `ResetWithRowSizes(row_sizes)`: size-driven construction with default-initialized data.
- `FromDataAndIndptr(data, indptr)`: move in a pre-built CSR.
- `MutableRowAt(i)`: in-place writable row.
- `Row::Slice(begin, end)`: sub-row view.

### 6. Cache `edge_num` on `CompactFSM` (supporting change)
`CompactFSMWithStartEnd::GetNumEdges()` becomes O(1). Bumps the serialization version `v11 → v12` because `CompactFSM` now serializes an extra field.

## Speedup
BFCL, Llama 3.1 8B, `tool_choice="required"`, `--repeats 1`, `max_threads=64`.

| tool_count | before | after | speedup |
|---:|---:|---:|---:|
| 100  | 501 ms     | 164 ms  | -67% |
| 200  | 1,312 ms   | 308 ms  | -77% |
| 500  | 3,971 ms   | 757 ms  | -81% |
| 1000 | 21,713 ms  | 2,085 ms | **-90%** |

Within the FSM stage at 1000 tools: removing `MinimizeDFA` saves ~580 ms, and `MergeEquivalentSuccessors` itself drops from ~230 ms to ~90 ms.

### DIFF
--- a/cpp/grammar_builder.cc
+++ b/cpp/grammar_builder.cc
@@ -262,13 +262,15 @@ void GrammarBuilder::UpdateLookaheadAssertion(
 std::string GrammarBuilder::GetNewRuleName(const std::string& name_hint) {
   if (rule_name_to_id_.count(name_hint) == 0) {
     return name_hint;
-  } else {
-    int cnt = 1;
-    while (rule_name_to_id_.count(name_hint + "_" + std::to_string(cnt)) != 0) {
-      ++cnt;
-    }
-    return name_hint + "_" + std::to_string(cnt);
   }
+  int* cnt = &next_cnt_per_hint_[name_hint];
+  if (*cnt == 0) {
+    *cnt = 1;
+  }
+  while (rule_name_to_id_.count(name_hint + "_" + std::to_string(*cnt)) != 0) {
+    ++(*cnt);
+  }
+  return name_hint + "_" + std::to_string(*cnt);
 }
 
 int32_t GrammarBuilder::GetRuleId(const std::string& name) const {
--- a/cpp/fsm.cc
+++ b/cpp/fsm.cc
@@ -569,7 +569,15 @@ class CompactFSM::Impl : public FSMImplBase<Compact2DArray<FSMEdge>> {
   using EdgeType = FSMEdge::EdgeType;
 
  public:
-  using FSMImplBase<Compact2DArray<FSMEdge>>::FSMImplBase;
+  Impl() = default;
+
+  Impl(const Compact2DArray<FSMEdge>& edges, std::vector<int32_t> edge_aux_data = {})
+      : FSMImplBase<Compact2DArray<FSMEdge>>(edges, std::move(edge_aux_data)),
+        edge_num_(ComputeEdgeNum(edges_)) {}
+
+  Impl(Compact2DArray<FSMEdge>&& edges, std::vector<int32_t> edge_aux_data = {})
+      : FSMImplBase<Compact2DArray<FSMEdge>>(std::move(edges), std::move(edge_aux_data)),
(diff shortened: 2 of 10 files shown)


cpp/grammar_builder.cc
```diff
@@ -262,13 +262,15 @@ void GrammarBuilder::UpdateLookaheadAssertion(
...
   if (rule_name_to_id_.count(name_hint) == 0) {
     return name_hint;
-  } else {
-    int cnt = 1;
-    while (rule_name_to_id_.count(name_hint + "_" + std::to_string(cnt)) != 0) {
-      ++cnt;
-    }
-    return name_hint + "_" + std::to_string(cnt);
   }
+  int& cnt = next_cnt_per_hint_[name_hint];
```
Seven-Streams: it's okay to use a reference, while using a pointer may be more readable.

PR #642, opened 2026-05-14
fix: escape dot in float boundary values for regex generation

### LINKED ISSUE #640: [Bug] `Grammar.from_json_schema` with float `minimum`/`maximum` produces wildcard instead of literal dot
### Checklist

### Describe the bug

When a JSON schema has a float `minimum` (e.g., `0.5`), `Grammar.from_json_schema` produces a grammar rule containing `[\0-\U0010ffff]` (match any character) where it should be `"."` (literal dot). This allows invalid outputs like `0,5`, `0;5`, etc.

```json
{
  "type": "number",
  "minimum": 0.5
}
```

- **Expected grammar rule**: `"0" "." "5"` (only matches `0.5`)
- **Actual grammar rule**: `"0" [\0-\U0010ffff] "5"` (matches `0<any char>5`, including `0,5`)

This affects both v0.1.25 and v0.2.0 — it is not a regression but a long-standing bug.

### Reproduction

```python
import xgrammar as xgr

schema = '{"type":"number","minimum":0.5}'
grammar = xgr.Grammar.from_json_schema(schema)
print(grammar)
# root ::= (("0" [\0-\U0010ffff] "5") | ("0" "." "6" ...) | ...)
#                 ^^^^^^^^^^^^^^^^ should be "."
```

Confirm by comparing regex behavior directly:

```python
# Escaped dot -> correct (literal dot)
g1 = xgr.Grammar.from_regex(r'0\.5')
print(g1)  # root ::= (("0" "." "5"))

# Unescaped dot -> bug (wildcard)
g2 = xgr.Grammar.from_regex('0.5')
print(g2)  # root ::= (("0" [\0-\U0010ffff] "5"))
```

### Runtime proof with real tokenizer

```python
import xgrammar as xgr

tok = AutoTokenizer.from_pretrained('tencent/Hy3-preview', trust_remote_code=True)
tokenizer_info = xgr.TokenizerInfo.from_huggingface(tok, vocab_size=tok.vocab_size)
compiler = xgr.GrammarCompiler(tokenizer_info)

schema = '{"type":"object","properties":{"score":{"type":"number","minimum":0.5}},"required":["score"]}'
ctx = compiler.compile_json_schema(schema)
matcher = xgr.GrammarMatcher(ctx)

# Feed: {"score": 0
for tid in tok.encode('{"score": 0'):
    matcher.accept_token(tid)

# Accept comma (should be rejected!)
comma_id = tok.encode(',')[0]
print(f"accept_token(',') after '0': {matcher.accept_token(comma_id)}")  # True -- BUG

# Accept '5'
five_id = tok.encode('5')[0]
print(f"accept_token('5') after ',': {matcher.accept_token(five_id)}")  # True

# Accept '}'
close_id = tok.encode('}')[0]
print(f"accept_token('}}') after '5': {matcher.accept_token(close_id)}")  # True

# Result: {"score": 0,5} accepted -- invalid JSON
```

Output:
```
accept_token(',') after '0': True    <-- BUG: comma should be rejected
accept_token('5') after ',': True
accept_token('}') after '5': True
```

### Root cause

In `cpp/json_schema_converter.cc`, `GenerateFloatRangeRegex` (line 2687) calls `FormatFloat(0.5)` which returns the string `"0.5"`. This string is directly pushed into the regex parts:

```cpp
// line 2720
std::string startIntStr = FormatFloat(start.value(), precision);
parts.push_back(startIntStr);  // "0.5" used as regex -- but '.' means "any char" in regex!
```

The `.` in `"0.5"` is interpreted as the regex wildcard (any character), not a literal dot. It should be escaped as `"0\\.5"`.

The same issue exists in other `parts.push_back(...)` calls within `GenerateFloatRangeRegex` where `FormatFloat` results are used -- any float value with a decimal point will have this bug.

Similarly, the `endIntStr` path (line 2776) and the `startStr`/`endStr` paths in the `(start && end)` branch have the same issue.

### Suggested fix

Add a helper to escape `.` in `FormatFloat` results before they are used as regex:

```cpp
static std::string EscapeForRegex(const std::string& s) {
  std::string result;
  for (char c : s) {
    if (c == '.') result += "\\.";
    else result += c;
  }
  return result;
}
```

Then replace the direct pushes:

```diff
  if (start && !end) {
    std::string startIntStr = FormatFloat(start.value(), precision);
-   parts.push_back(startIntStr);
+   parts.push_back(EscapeForRegex(startIntStr));
```

```diff
  } else if (!start && end) {
    std::string endIntStr = FormatFloat(end.value(), precision);
-   parts.push_back(endIntStr);
+   parts.push_back(EscapeForRegex(endIntStr));
```

And similarly for `startStr`/`endStr` in the `(start && end)` branch.

Note: other `parts.push_back(...)` calls in this function already use `"\\."` explicitly (e.g., `intPartStr + "\\." + d + ...`), so they are correct. The bug is only where `FormatFloat` results are pushed directly without dot escaping.

### Environment

- xgrammar version: v0.1.25 and v0.2.0 (both affected, `main` at commit `41dbbb18`)
- Python: 3.13
- OS: macOS (arm64)

dynamicheart:
## Summary

`FormatFloat(0.5)` returns `"0.5"` which was pushed directly into regex parts in `GenerateFloatRangeRegex`. The unescaped `.` was interpreted as the regex wildcard (match any character), producing grammar rules like:

```
root ::= (("0" [\0-\U0010ffff] "5") | ...)
                ^^^^^^^^^^^^^^^^ should be "."
```

This allowed invalid outputs like `0,5` (European decimal) at runtime, since any character was accepted in place of the decimal point.

This is a long-standing bug affecting both v0.1.25 and v0.2.0.

## Root Cause

In `GenerateFloatRangeRegex`, boundary values from `FormatFloat` are pushed directly into regex `parts` without escaping the `.` character. Other parts of the same function already use `"\\."` explicitly (e.g., `intPartStr + "\\." + d + ...`), but the boundary value strings themselves were never escaped.

## Changes

- Add `EscapeDotForRegex()` static helper that replaces `.` with `\\.`
- Apply it to all 6 `parts.push_back(FormatFloat(...))` calls in `GenerateFloatRangeRegex`
- Update `test_generate_float_regex` expected values to use escaped dots
- Add `test_float_minimum_no_wildcard_in_grammar` end-to-end test

## Reproduction

```python
import xgrammar as xgr

# Before fix: "0" [\0-\U0010ffff] "5" (wildcard)
# After fix:  "0" "." "5" (literal dot)
schema = '{"type":"number","minimum":0.5}'
grammar = xgr.Grammar.from_json_schema(schema)
print(grammar)
```

Confirm with regex API:
```python
# Escaped dot → correct
g1 = xgr.Grammar.from_regex(r'0\.5')   # root ::= (("0" "." "5"))

# Unescaped dot → bug (before fix)
g2 = xgr.Grammar.from_regex('0.5')     # root ::= (("0" [\0-\U0010ffff] "5"))
```

## Test

```bash
python3 -m pytest tests/python/test_json_schema_converter.py::test_generate_float_regex tests/python/test_json_schema_converter.py::test_float_minimum_no_wildcard_in_grammar -v
# 2 passed

python3 -m pytest tests/python/test_json_schema_converter.py -v
# 320 passed
```

Fixes #640

cc @Seven-Streams

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -2684,6 +2684,19 @@ std::string JSONSchemaConverter::FormatFloat(double value, int precision) {
   return result;
 }
 
+static std::string EscapeDotForRegex(const std::string& s) {
+  std::string result;
+  result.reserve(s.size() + 2);
+  for (char c : s) {
+    if (c == '.') {
+      result += "\\.";
+    } else {
+      result += c;
+    }
+  }
+  return result;
+}
+
 std::string JSONSchemaConverter::GenerateFloatRangeRegex(
     std::optional<double> start, std::optional<double> end, int precision
 ) {
@@ -2718,7 +2731,7 @@ std::string JSONSchemaConverter::GenerateFloatRangeRegex(
 
   if (start && !end) {
     std::string startIntStr = FormatFloat(start.value(), precision);
-    parts.push_back(startIntStr);
+    parts.push_back(EscapeDotForRegex(startIntStr));
 
     if (startFrac > 0.0) {
       size_t dotPos = startIntStr.find('.');
@@ -2774,7 +2787,7 @@ std::string JSONSchemaConverter::GenerateFloatRangeRegex(
     }
   } else if (!start && end) {
     std::string endIntStr = FormatFloat(end.value(), precision);
-    parts.push_back(endIntStr);
+    parts.push_back(EscapeDotForRegex(endIntStr));
 
     if (endFrac > 0.0) {
       size_t dotPos = endIntStr.find('.');
@@ -2833,20 +2846,20 @@ std::string JSONSchemaConverter::GenerateFloatRangeRegex(
         parts.push_back(std::to_string(startInt));
       } else {
         std::string startStr = FormatFloat(start.value(), precision);
-        parts.push_back(startStr);
+        parts.push_back(EscapeDotForRegex(startStr));
 
         std::string endStr = FormatFloat(end.value(), precision);
         if (startStr != endStr) {
-          parts.push_back(endStr);
+          parts.push_back(EscapeDotForRegex(endStr));
         }
       }
     } else {
       std::string startStr = FormatFloat(start.value(), precision);
-      parts.push_back(startStr);
+      parts.push_back(EscapeDotForRegex(startStr));
 
       std::string endStr = FormatFloat(end.value(), precision);
       if (startStr != endStr) {
-        parts.push_back(endStr);
+        parts.push_back(EscapeDotForRegex(endStr));
       }
 
       if (endInt > startInt + 1) {
--- a/tests/python/test_json_schema_converter.py
+++ b/tests/python/test_json_schema_converter.py
@@ -2158,36 +2158,59 @@ def test_generate_float_regex():
 
     assert (
         _generate_float_regex(1.5, 5.75)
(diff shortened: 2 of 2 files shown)


Seven-Streams (approved):
Thanks, LGTM!

PR #645, opened 2026-05-26
fix: replace unsupported inline style with SVG for WebLLM logo in README

1195343015:
GitHub strips inline CSS style attributes from markdown, so the span element with font-size did not render. Replace it with an inline SVG that achieves the same visual effect and works on GitHub.


Ubospica:
@1195343015 Thanks for the contribution! However, it seems that the fix does not work. It would be great we can find other method to fix this.

PR #648, opened 2026-05-29
fix: store FSMEdge rule_id as unsigned 16-bit to support >32767 rules

fsaudm:
## Summary

`FSMEdge` packs a `kRuleRef` edge's `rule_id` into the signed `int16_t max` field, and `FSMEdge::GetRefRuleId()` returned it sign-extended. Once an optimized grammar has more than **32,767** rules, edges that reference higher rule indices wrap to negative values (e.g. rule_id 33,493 is read back as -32,043). `EarleyParser::ExpandNextRuleRefElementOnFSM` then indexes `per_rule_fsms[ref_rule_id]` with the negative value, dereferences a default-constructed `std::optional`, and throws `std::bad_optional_access`, a hard crash (segfault in release builds). `RepeatEdgeRef::RuleId()` had the same truncation even though its aux storage is already `int32_t`.

## Reproduction

```python
import xgrammar as xgr

N = 33000   # > 32767 rules after optimization
lines = ["root ::= sub", "sub ::= " + " | ".join(f"p{i}" for i in range(N))]
for i in range(N):
    lines.append(f'p{i} ::= "a" (p{i+1} | "")' if i < N - 1 else f'p{i} ::= "a"')
gbnf = "\n".join(lines)

xgr.GrammarCompiler(tinfo).compile_grammar(gbnf)   # std::bad_optional_access
```

The threshold is exact: 32,768 rules compiles, 32,770 crashes.

## Fix

Treat the 16-bit storage slot as **unsigned**, extending the valid range from `[0, 32767]` to `[0, 65535]`:

- `FSMEdge::GetRefRuleId()` returns `static_cast<int32_t>(static_cast<uint16_t>(max))`.
- `RepeatEdgeRef::RuleId()` returns the full `int32_t` from its aux storage.
- `FSM::AddRuleEdge` takes `int32_t rule_id` and `XGRAMMAR_CHECK`s `0 <= rule_id <= 0xFFFF`, so grammars beyond 65,535 rules now fail with a clear message instead of silently corrupting.
- The one call site that pre-truncated with `static_cast<int16_t>` is updated accordingly.

Storage layout, struct size, and on-disk serialization are unchanged — only the interpretation of the `max` field widens from signed to unsigned.

## Risk

- Grammars with at most 32,767 rules: bit-identical behavior.
- Grammars with 32,768–65,535 rules: previously crashed, now correct.
- Grammars beyond 65,535 rules: previously crashed silently, now produce a clear `XGRAMMAR_CHECK` at FSM build time.
- No upstream call site reads `FSMEdge::max` directly (bypassing `GetRefRuleId()`) for `kRuleRef` edges, so the signedness change is contained.

## Notes

Independent of the companion `GrammarFSMHasher` fix, though both surface on very large grammars and are best applied together.

### DIFF
--- a/cpp/fsm.cc
+++ b/cpp/fsm.cc
@@ -107,7 +107,10 @@ std::string FSMImplBase<ContainerType>::EdgesToString(std::optional<std::vector<
         std::string char_str = EscapeString(static_cast<TCodepoint>(edge.min));
         result += "'" + char_str + "'->" + std::to_string(edge.target);
       } else if (edge.min == FSMEdge::EdgeType::kRuleRef) {
-        result += "Rule(" + std::to_string(edge.max) + ")->" + std::to_string(edge.target);
+        // Print the unsigned rule_id (GetRefRuleId), not the raw signed `max`,
+        // so rule_ids > 32767 are shown correctly instead of as negatives.
+        result +=
+            "Rule(" + std::to_string(edge.GetRefRuleId()) + ")->" + std::to_string(edge.target);
       } else if (edge.min == FSMEdge::EdgeType::kEpsilon) {
         result += "Eps->" + std::to_string(edge.target);
       } else if (edge.min == FSMEdge::EdgeType::kEOS) {
@@ -246,8 +249,21 @@ class FSM::Impl : public FSMImplBase<std::vector<std::vector<FSMEdge>>> {
     edges_[from].push_back({min, max, to});
   }
 
-  void AddRuleEdge(int from, int to, int16_t rule_id) {
-    AddEdge(from, to, FSMEdge::EdgeType::kRuleRef, rule_id);
+  void AddRuleEdge(int from, int to, int32_t rule_id) {
+    // rule_id is stored in the int16_t `max` field of FSMEdge, but interpreted
+    // as an unsigned 16-bit value by FSMEdge::GetRefRuleId(). That gives a
+    // valid range of [0, 65535]. Reject anything outside the range with a
+    // clear error message instead of silently truncating (which previously
+    // wrapped rule_ids > 32767 to negatives and caused per_rule_fsms[i]
+    // out-of-range crashes in EarleyParser at compile time).
+    XGRAMMAR_CHECK(rule_id >= 0 && rule_id <= 0xFFFF)
+        << "AddRuleEdge: rule_id " << rule_id
+        << " is out of range [0, 65535]. The xgrammar FSM stores rule_ids in "
+           "a 16-bit field; grammars with more than 65536 rules are not "
+           "supported by this build.";
+    AddEdge(
+        from, to, FSMEdge::EdgeType::kRuleRef, static_cast<int16_t>(static_cast<uint16_t>(rule_id))
+    );
   }
 
   void AddEpsilonEdge(int from, int to) { AddEdge(from, to, FSMEdge::EdgeType::kEpsilon, 0); }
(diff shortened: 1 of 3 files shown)


fsaudm:
Companion to #647: both crashes surface on very large right-recursive grammars. This PR fixes the `int16_t` rule_id overflow (>32767 rules); #647 fixes a `GrammarFSMHasher` assertion that fires earlier (~5k+ rules). Independent fixes, best reviewed together. :  )

fsaudm:
Good catch @gemini-code-assist (funny talking with an AI agent). Fixed in 37bf08d: the three rule_id comparison sites that read the raw signed `edge.max` now compare via `edge.GetRefRuleId()` (unsigned 16-bit read):

- `FSM::Impl::GetNextState`
- `CompactFSM::Impl::GetNextStates`
- `CompactFSM::Impl::Advance`

`FSM::Impl::Advance` already used `GetRefRuleId()` (no change needed), and `EdgesToString` now prints `GetRefRuleId()` so high rule_ids aren't shown as negatives.

One note for reviewers on severity: these three methods are not on the Earley matcher's hot path;`ExpandNextRuleRefElementOnFSM`/`Predict` discover rule refs by iterating a state's edges and reading `GetRefRuleId()` directly, never by searching for a rule by value through `GetNextState`/`GetNextStates`/`Advance(kRuleRef)` (which have no kRuleRef callers in `cpp/` today). So for decoding the load-bearing fix is the getter; these comparison-site fixes harden the public FSM API and remove the latent correctness gap for any other/future consumer. Verified with a small FSM unit test using a `Rule(40000)` edge: all four traversal methods now return the target state (pre-fix, `GetNextState` returned -1 and the CompactFSM methods returned empty).

Ubospica:
@fsaudm Thanks for the fix! The fix makes sense to me.

Previously, we set this to 16 bit because there are very few cases we observed where the number of rules is more than 30k, and 16bit datatype better packs into CPU's cache line. I'm just wondering if there are any significant use cases that require more rules than that. If so, we can consider further expanding the datatype.

Ubospica:
@Seven-Streams Let's check if we can use int32 and what's the efficiency consequences. If the impact is not much, we can switch to int32. Seems #647 could also be caused by int16 overflow of the rule id. With overflow, a edge may refer to a wrong rule id, causing the cross-grammar cache to fail.

fsaudm:
Thanks @Ubospica! I am using xgrammar with vLLM for "grounded extraction" via structured_output for guided decoding ([https://docs.vllm.ai/en/latest/features/structured_outputs/](https://docs.vllm.ai/en/latest/features/structured_outputs/)), where I constrain the LLM to emit an `evidence` field that is a byte-exact substring of the source doc (model can't hallucinate, Im forcing it to reproduce what it saw as input!). GBNF is character-level, so I encode "any substring of D" as a chain with one rule per character, and a ~22k-char doc becomes ~44k rules. It is document length that blows past 16 bit, so probably not a super common shape.

There are ways to avoid >32k rules entirely (coarsen the chain unit from char to word, or skip the grammar and use a custom logits processor), so I am not blocked either way. But if int32 turns out cheap on the hot path, it would be nice to just have it work, and happy to help @Seven-Streams benchmark it : )

Either way I think this PR is still worth it just for turning the silent overflow into a clean error

Seven-Streams:
@fsaudm Sorry for the late response. It seems that it will not hurt the efficiency, and I've opened a pr here #652.

fsaudm:
Thanks @Seven-Streams! #652 looks great (and I really appreciate you building it on top of my commit). I'm building #652 now to run it against my "real" workload that originally hit the overflow (a ~142k-rule substring grammar from a long doc) and will report results over there.

Issue #489, opened 2025-12-15
[Bug] Relax XML Parameter Validation to Allow Arbitrary Content in <parameter=content> Tags


Ssskrilex:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. I have filled all the required information below.
- [x] 4. The bug report is in English.

## Description
The current validation rules for XML-formatted function call parameters (`<parameter=content>...</parameter>`) are overly restrictive and contradict the core purpose of this syntax: handling escape issues for unstructured/structured text (code, multi-line strings, HTML/XML, etc.) in JSON payloads.

### Current Issue
Validation currently blocks valid use cases by restricting the content inside `<parameter=content>` tags. The intended design of this XML syntax is to act as a "wrapper" for arbitrary text that would otherwise require complex escaping in JSON – but strict content validation negates this purpose.

### Expected Behavior
The validation logic should:
1. **Only validate the tag structure**: Ensure the opening `<parameter=content>` and closing `</parameter>` tags are present and properly formatted.
2. **Ignore content validation entirely**: The content inside the tags should be treated as a raw string (the entire point of this XML wrapper syntax).
---

### Examples of Valid Content Being Rejected
These should all be accepted (but are currently blocked):
```xml
<!-- Multi-line code with newlines/quotes -->
<parameter=content>import pandas

print("test: 123")
df = pd.read_csv("data.csv")</parameter>

<!-- HTML/XML markup (including doctypes) -->
<parameter=content><!DOCTYPE html>
<html lang="en">
  <body><h1>Hello</h1></body>
</html></parameter>

<!-- Unstructured text with special characters -->
<parameter=content>User input: !@#$%^&*()_+-=[]{}|;':",./<>?
Multi-line
content with mixed symbols and whitespace</parameter>
```

### Environment

https://github.com/mlc-ai/xgrammar/pull/387
cc @Seven-Streams

Seven-Streams:
Hello, thanks for your report! I'm not sure if I got your point. You mean that if there is no constraint on the parameter, then we should allow all the tokens until the end tag is output. Is that correct?

Ssskrilex:
Yes, for string type parameter.

Seven-Streams:
I agree that some characters should be accepted, such as `\n` and `\r`. However, according to [XML](https://www.w3.org/TR/xml/):

[...]

PR #651, opened 2026-06-05
Lock excluded-string prefixes in TagDispatch for XML parameter values

### LINKED ISSUE #650: [Bug] XML parameter values can drift past the end-marker prefix under constrained decoding (e.g. `</parameter1>`)
### Summary

For XML tool-call formats (`qwen_xml` / `minimax_xml` / `deepseek_xml` / `glm_xml`), a `string`-typed parameter value is constrained as:

```
xml_string ::= TagDispatch(loop_after_dispatch=false, excludes=("</parameter>"))
```

i.e. "any content until the end marker `</parameter>`". Under **constrained decoding**, once the model has emitted the longest *prefix* of the end marker (`</parameter`), the grammar does **not** force the next character to be `>`. It lets `</parameter` fall back into ordinary content, so the model can emit a malformed end marker such as `</parameter1>`, `</parameter_function>`, `</parameter/>`. The consequences:

1. The malformed marker is not a valid `</parameter>`, so the value never terminates — the model keeps generating until the length limit.
2. If generation is truncated / loosely re-parsed downstream, the malformed marker leaks into the parsed arguments, e.g. `{"city": "上海\n</parameter1>"}`.

Originally reported downstream at sgl-project/sglang#27336 (Qwen3.6 / `qwen3_coder` with `tool_choice="required"`).

### Reproduction

```python
import xgrammar as xg
import xgrammar.testing as t

schema = {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}
g = xg.Grammar.from_ebnf(t._qwen_xml_tool_calling_to_ebnf(schema))

# A malformed end marker is (wrongly) a valid continuation:
print(t._is_grammar_accept_string(g, "<parameter=city>Shanghai</parameter1", require_termination=False))
# -> True   (the grammar does not force '>' after '</parameter')

# It even swallows a malformed marker as content while still waiting for the real end:
print(t._is_grammar_accept_string(g, "<parameter=city>Shanghai</parameter1>more</parameter>", require_termination=True))
# -> True
```

### Root cause

`TagDispatch(excludes=...)` is compiled (`cpp/fsm_builder.cc`, `TrieFSMBuilderImpl::Build` + `AddBackEdges`) into an Aho-Corasick automaton: a trie of the excluded string plus failure/back edges. Every **non-terminal** trie state — including the longest prefix state `</parameter` — receives back edges that send any non-matching character back to the start, i.e. "treat what we matched so far as ordinary content and keep going". The edge into the excluded terminal state is then removed so the excluded string itself can never be emitted as content.

That back-edge-at-every-prefix behaviour is correct for the general "any text until a delimiter" use (the delimiter prefix may legitimately appear inside the content). But for a **constrained-decoding token mask** it means: at `</parameter`, the mask permits both `>` (complete & stop) *and* any other byte (fall back to content), so the model is free to drift into `</parameter1>`.

This is a regression from #511 (relaxing XML string values to allow arbitrary content incl. `<`, per #489). Before #511 the value was `[^<>&...]`-based, so `<` could not appear in content and the end marker was effectively forced; #511 switched it to the TagDispatch form to support `<`/HTML, which introduced the drift. Confirmed present on both 0.2.0 and 0.2.1 (latest release).

### Proposed fix

Add an opt-in `lock_excluded_prefixes` flag to `TagDispatch`. When set, the **longest proper prefix** of each excluded string gets no fall-back back edges: from `</parameter` the only options are to complete `</parameter>` (stop) or restart a fresh match on a pattern-initial char (`<`). Shorter prefixes keep their back edges, so legitimate `<`-containing content (`<div>`, `</p>`, `1 < 2 & 3 > 0`) is still accepted; only a value that contains the exact full prefix `</parameter` is rejected — an acceptable trade-off. Enable it for the XML parameter-value rule only; reasoning / `any_text` TagDispatch keeps the existing "any content until delimiter" semantics.

PR: https://github.com/mlc-ai/xgrammar/pull/651

JustinTong0323:
## Summary

Fix XML tool-call string parameter values drifting past the end-marker prefix under constrained decoding. Closes #650 (reported downstream at sgl-project/sglang#27336).

For `qwen_xml` / `minimax_xml` / `deepseek_xml` / `glm_xml`, a `string` parameter value is compiled as `TagDispatch(excludes=("</parameter>"))` — "any content until the end marker". The Aho-Corasick back edges added for the excluded string let the value drift past the longest end-marker prefix (`</parameter`) back into ordinary content. Under constrained decoding the token mask at `</parameter` therefore allows any byte (not just `>`), so the model can emit a malformed end marker like `</parameter1>` / `</parameter_function>` and then either leak it into the parsed arguments or keep generating to the length limit.

This is a regression from #511 (which relaxed XML string values to allow `<`/HTML per #489): before #511 the value forbade `<`, so the end marker was effectively forced.

## Modifications

- `TagDispatch` gains a `lock_excluded_prefixes` option (struct field + EBNF named arg + parser/printer + builder encoding). Default `false` preserves current behaviour.
- `fsm_builder.cc`: when the option is set, the **longest proper prefix** state of each excluded string is excluded from `AddBackEdges`, so it cannot fall back to content — it must complete the excluded string (or restart a fresh match on a pattern-initial char, which keeps shorter prefixes / legitimate `<`-content working).
- `json_schema_converter_ext.cc`: enable `lock_excluded_prefixes=true` for the XML parameter-value rule (`kXMLString`). All four XML formats benefit; reasoning / `any_text` TagDispatch is untouched.

## Behaviour

For a `{"city": string}` schema:

| Value emitted | Before | After |
|---|---|---|
| `Shanghai</parameter>` | accept | accept |
| `<div>x</div></parameter>` (HTML) | accept | accept |
| `1 < 2 & 3 > 0</parameter>` | accept | accept |
| `Shanghai</parameter1>` | **accept (drift)** | **reject** |
| `Shanghai</parameter_function>` | **accept (drift)** | **reject** |

Reasoning `any_text ::= TagDispatch(excludes=("</think>"))` is unchanged (still accepts arbitrary content containing prefixes of the delimiter).

## Tests

- `tests/python/test_function_calling_converter.py`: new `test_string_end_marker_lock` (malformed markers rejected, HTML / `<`-content accepted); existing expected-grammar strings updated to include the new flag.
- `tests/python/test_structural_tag_converter.py`: expected XML grammars updated; `test_any_text_format` (and the rest of the suite) confirms reasoning/any_text semantics are preserved.
- Full local run: `test_function_calling_converter.py`, `test_structural_tag_converter.py`, `test_builtin_structural_tag.py` → 1326 passed.

### DIFF
--- a/cpp/fsm_builder.cc
+++ b/cpp/fsm_builder.cc
@@ -784,9 +784,15 @@ class TrieFSMBuilderImpl {
       const std::vector<std::string>& excluded_patterns,
       std::vector<int32_t>* end_states,
       bool allow_overlap,
-      bool add_back_edges
+      bool add_back_edges,
+      bool lock_excluded_prefixes
   );
-  void AddBackEdges(FSM* fsm, int start, const std::unordered_set<int>& ends);
+  void AddBackEdges(
+      FSM* fsm,
+      int start,
+      const std::unordered_set<int>& ends,
+      const std::unordered_set<int>& locked_states
+  );
 };
 
 std::optional<FSMWithStartEnd> TrieFSMBuilderImpl::Build(
@@ -794,11 +800,16 @@ std::optional<FSMWithStartEnd> TrieFSMBuilderImpl::Build(
     const std::vector<std::string>& excluded_patterns,
     std::vector<int32_t>* end_states,
     bool allow_overlap,
-    bool add_back_edges
+    bool add_back_edges,
+    bool lock_excluded_prefixes
 ) {
   FSM fsm(1);
   int start = 0;
   std::unordered_set<int> ends;
+  // The longest proper prefix state of each excluded pattern. When lock_excluded_prefixes is set,
+  // these states get no back edges so the prefix can only be completed into the full excluded
+  // pattern (see header doc).
+  std::unordered_set<int> locked_states;
 
   if (end_states) {
     end_states->clear();
@@ -842,6 +853,7 @@ std::optional<FSMWithStartEnd> TrieFSMBuilderImpl::Build(
       }
 
       int current_state = 0;
+      int penultimate_state = start;
       for (const auto& ch : excluded_pattern) {
         int16_t ch_int16 = static_cast<int16_t>(static_cast<uint8_t>(ch));
         int next_state = fsm.GetNextState(current_state, ch_int16);
@@ -849,6 +861,7 @@ std::optional<FSMWithStartEnd> TrieFSMBuilderImpl::Build(
           next_state = fsm.AddState();
           fsm.AddEdge(current_state, next_state, ch_int16, ch_int16);
         }
+        penultimate_state = current_state;
         current_state = next_state;
         if (!allow_overlap && ends.count(current_state) > 0) {
           return std::nullopt;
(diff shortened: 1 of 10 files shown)


Seven-Streams (approved):
That's indeed a potential problem, thanks a lot!

Swipe4057:
@Ubospica Hi! Could you please review the changes? sglang has currently switched to structural tags without a full-back to the old processing logic, and several versions of sglang are already affected by this bug, making updates difficult in production systems.

PR #652, opened 2026-06-06
fix: store FSMEdge rule_id as int_32 to the cases with >32767 rules.

Seven-Streams:
This pr changes use int32 to store edges instead of int16, which supports cases with more than 32767 rules.

The benchmark code is as follows:

```python
"""Micro-benchmark for the FSM edge int16->int32 change.

Test point: `MainModel` JSON schema taken from
tests/python/test_grammar_matcher_json_schema.py (covers ints, floats, bools,
arrays, tuples, nested objects). Uses the non-gated gpt2 tokenizer so it runs
without a HF token / GPU.

Measures:
  * compilation time   : GrammarCompiler.compile_json_schema (fresh compiler/iter)
  * mask generation    : GrammarMatcher.fill_next_token_bitmask (CPU; this is the
                         FSM path that the edge struct change affects)
"""

import argparse
import statistics
import time
from typing import List, Tuple

from pydantic import BaseModel

import xgrammar as xgr


class MainModel(BaseModel):
    integer_field: int
    number_field: float
    boolean_field: bool
    any_array_field: list
    array_field: List[str]
    tuple_field: Tuple[str, int, List[str]]
    object_field: dict
    nested_object_field: dict


instance = MainModel(
    integer_field=42,
    number_field=3.14e5,
    boolean_field=True,
    any_array_field=[3.14, "foo", None, True],
    array_field=["foo", "bar"],
    tuple_field=("foo", 42, ["bar", "baz"]),
    object_field={"foo": 42, "bar": 43},
    nested_object_field={"foo": {"bar": 42}},
)
INSTANCE_STR = instance.model_dump_json(indent=2)


def bench(tokenizer_name: str, num_iters: int, num_warmup: int):
    tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
    tokenizer_info = xgr.TokenizerInfo.from_huggingface(tokenizer)
    vocab_size = tokenizer_info.vocab_size

    compile_times_ms: List[float] = []
    mask_times_us: List[float] = []
    num_mask_calls = len(INSTANCE_STR)

    for it in range(-num_warmup, num_iters):
        # --- compilation time: fresh compiler each iter to avoid cache hits ---
        compiler = xgr.GrammarCompiler(tokenizer_info, cache_enabled=False)
        t0 = time.perf_counter()
        compiled_grammar = compiler.compile_json_schema(MainModel, indent=2)
        t1 = time.perf_counter()

        # --- mask generation time: one fill_next_token_bitmask per char ---
        matcher = xgr.GrammarMatcher(compiled_grammar)
        bitmask = xgr.allocate_token_bitmask(1, vocab_size)
        t2 = time.perf_counter()
        for c in INSTANCE_STR:
            matcher.fill_next_token_bitmask(bitmask)
            assert matcher.accept_string(c), f"rejected char {c!r}"
        t3 = time.perf_counter()

        if it >= 0:
            compile_times_ms.append((t1 - t0) * 1e3)
            mask_times_us.append((t3 - t2) / num_mask_calls * 1e6)

    def summarize(xs):
        return statistics.mean(xs), statistics.median(xs), min(xs)

    c_mean, c_med, c_min = summarize(compile_times_ms)
    m_mean, m_med, m_min = summarize(mask_times_us)

    print(f"xgrammar serialize version : {xgr.get_serialization_version()}")
    print(f"tokenizer                  : {tokenizer_name} (vocab={vocab_size})")
    print(f"iters/warmup               : {num_iters}/{num_warmup}")
    print(f"mask calls per iter        : {num_mask_calls}")
    print("-" * 56)
    print(f"compilation time   (ms)    : mean={c_mean:.4f}  median={c_med:.4f}  min={c_min:.4f}")
    print(f"mask gen time  (us/token)  : mean={m_mean:.4f}  median={m_med:.4f}  min={m_min:.4f}")


if __name__ == "__main__":
    ap = argparse.ArgumentParser()
    ap.add_argument("--tokenizer", default="meta-llama/Llama-3.1-8B-Instruct")
    ap.add_argument("--num_iters", type=int, default=200)
    ap.add_argument("--num_warmup", type=int, default=20)
    args = ap.parse_args()
    bench(args.tokenizer, args.num_iters, args.num_warmup)
```

and the results are:

```text
int_32
xgrammar serialize version: v14
tokenizer                  : meta-llama/Llama-3.1-8B-Instruct (vocab=128256)
iters/warmup               : 200/20
mask calls per iter        : 392
--------------------------------------------------------
compilation time   (ms)    : mean=16.8092  median=16.9765  min=13.1094
mask gen time  (us/token)  : mean=16.6398  median=17.0990  min=9.3919

int_16
xgrammar serialize version: v13
tokenizer                  : meta-llama/Llama-3.1-8B-Instruct (vocab=128256)
iters/warmup               : 200/20
mask calls per iter        : 392
--------------------------------------------------------
compilation time   (ms)    : mean=16.8140  median=16.7889  min=13.3949
mask gen time  (us/token)  : mean=15.2320  median=16.3241  min=9.2153
```

The result demonstrates that it will not hurt the efficiency.

### DIFF
--- a/cpp/support/json_serializer.h
+++ b/cpp/support/json_serializer.h
@@ -62,7 +62,7 @@ class SerializeVersion {
    * \brief The current serialization version. When the serialization result of any object in
    * XGrammar is changed, this version should be bumped.
    */
-  static constexpr const char kXGrammarSerializeVersion[] = "v14";
+  static constexpr const char kXGrammarSerializeVersion[] = "v13";
 };
 
 /*!
--- a/cpp/support/json_serializer.h
+++ b/cpp/support/json_serializer.h
@@ -62,7 +62,7 @@ class SerializeVersion {
    * \brief The current serialization version. When the serialization result of any object in
    * XGrammar is changed, this version should be bumped.
    */
-  static constexpr const char kXGrammarSerializeVersion[] = "v13";
+  static constexpr const char kXGrammarSerializeVersion[] = "v14";
 };
 
 /*!
--- a/cpp/fsm.cc
+++ b/cpp/fsm.cc
@@ -76,11 +76,11 @@ class FSMImplBase {
 
   void SetEdgeAuxData(std::vector<int32_t> data) { edge_aux_data_ = std::move(data); }
 
-  RepeatEdgeRef GetRepeatEdgeInfo(int16_t idx) const { return {edge_aux_data_.data() + idx}; }
+  RepeatEdgeRef GetRepeatEdgeInfo(int32_t idx) const { return {edge_aux_data_.data() + idx}; }
 
-  TokenEdgeRef GetTokenEdgeInfo(int16_t idx) const { return {edge_aux_data_.data() + idx}; }
+  TokenEdgeRef GetTokenEdgeInfo(int32_t idx) const { return {edge_aux_data_.data() + idx}; }
 
-  ExcludeTokenEdgeRef GetExcludeTokenEdgeInfo(int16_t idx) const {
+  ExcludeTokenEdgeRef GetExcludeTokenEdgeInfo(int32_t idx) const {
     return {edge_aux_data_.data() + idx};
   }
 
@@ -241,12 +241,12 @@ class FSM::Impl : public FSMImplBase<std::vector<std::vector<FSMEdge>>> {
     return edges_.size() - 1;
   }
 
-  void AddEdge(int from, int to, int16_t min, int16_t max) {
+  void AddEdge(int from, int to, int32_t min, int32_t max) {
     XGRAMMAR_DCHECK(from < static_cast<int>(edges_.size()));
     edges_[from].push_back({min, max, to});
   }
 
-  void AddRuleEdge(int from, int to, int16_t rule_id) {
+  void AddRuleEdge(int from, int to, int32_t rule_id) {
     AddEdge(from, to, FSMEdge::EdgeType::kRuleRef, rule_id);
   }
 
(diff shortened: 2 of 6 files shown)


fsaudm:
Following up from #648: I built this branch from source (macOS arm64, Python 3.14) and ran it against the workload that originally hit the overflow: the char-level substring grammar, same setup as the repro in #648.

```python
import time, string
import xgrammar as xgr

N = 40_000  # or 100_000

chars = string.ascii_letters + string.digits + " "
text = (chars * (N // len(chars) + 1))[:N - 2]   # ~N rules, one per char position

# real doc (~142k chars):
# from markitdown import MarkItDown
# text = MarkItDown().convert("https://www.sec.gov/files/litigation/admin/2016/34-78989.pdf").text_content

def esc(c):  # GBNF-escape a char for use inside a "..." literal
    return {'"': '\\"', '\\': '\\\\', '\n': '\\n', '\r': '\\r', '\t': '\\t'}.get(c, c)

n = len(text)
rules = [
    f'p{i} ::= "{esc(text[i])}" (p{i+1} | "")' if i < n - 1 else f'p{i} ::= "{esc(text[i])}"'
    for i in range(n)
]
gbnf = "\n".join(
    ["root ::= substring", "substring ::= " + " | ".join(f"p{i}" for i in range(n)), *rules]
)

# gpt2 tokenizer: ungated + what xgrammar's own tests/benchmarks use. 
# Any tokenizer would works (I use gemma in production); vocab size would/could the compile time.
t0 = time.time()
xgr.GrammarCompiler(tinfo).compile_grammar(xgr.Grammar.from_ebnf(gbnf))
print(f"compiled OK: ~{n + 2} rules in {time.time() - t0:.1f}s")
```

and the results are:

| rules | compile time | peak RSS | result |
|---|---|---|---|
| 40,000 | 26.3s | 3.2 GB | OK |
| 100,000 | 170.5s | 8.7 GB | OK |
| 142,092 | 3582.6s | 8.7 GB | OK |

For reference, the 40k case crashes on any unfixed build. The 142k-rule case is the same builder but with `text` set to the longest real doc in my dataset, and it now compiles cleanly. At large N the cost is dominated by grammar compilation itself, not the edge width.

Thanks again @Seven-Streams! : )

cpp/support/json_serializer.h
```diff
@@ -62,7 +62,7 @@ class SerializeVersion {
    * \brief The current serialization version. When the serialization result of any object in
    * XGrammar is changed, this version should be bumped.
    */
-  static constexpr const char kXGrammarSerializeVersion[] = "v13";
+  static constexpr const char kXGrammarSerializeVersion[] = "v14";
```
Ubospica:
```suggestion
  static constexpr const char kXGrammarSerializeVersion[] = "v13";
```

Ubospica (approved):

Ubospica:
Thanks @Seven-Streams , and @fsaudm for testing!

This should also resolve #648 and #647.

PR #661, opened 2026-06-15
fix: correct unsound grammars for number/integer range keywords

igt0-jota:
GenerateFloatRangeRegex produced grammars that were wrong in both directions for "type": "number" schemas with range bounds:

- For integer-valued bounds the fractional patterns were never generated (guarded by startFrac > 0.0), so for {"exclusiveMinimum": 0} every value in (0, 1) was unrepresentable. Under token-masked decoding a model writing 0.1 is silently coerced to 0.
- Exclusive bounds were treated as inclusive: exclusiveMinimum / exclusiveMaximum compiled identically to minimum / maximum, so the excluded bound itself was accepted.
- The unbounded integer alternative used [1-9]\d*, which matches values below the bound ({"minimum": 2} accepted 1). This also affected "type": "integer" schemas via GenerateRangeRegex.
- The both-bounds branch emitted endInt\.\d{1,6}, accepting values above the maximum ([1, 5] accepted 5.7).
- GenerateRangeRegex with only a negative minimum produced an empty subrange ({"type": "integer", "minimum": -5} matched no negative integers at all, only the literal "-").

Rewrite GenerateFloatRangeRegex with a digit-by-digit enumeration of the boundary fraction families (lower bound, upper bound, and two-sided same-integer-part ranges), and plumb exclusive_start / exclusive_end flags from NumberSpec through to the regex generator. Fix the two broken branches in GenerateRangeRegex.

Tests: value-level acceptance tests for number and integer range keywords, plus a sweep comparing grammar acceptance against plain float comparison around every bound; expected-regex tests updated to the corrected generator output (verified value-level first).

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -1438,21 +1438,29 @@ std::string JSONSchemaConverter::GenerateNumber(
     const NumberSpec& spec, const std::string& rule_name
 ) {
   std::optional<double> start, end;
+  bool exclusive_start = false;
+  bool exclusive_end = false;
   if (spec.minimum.has_value()) {
     start = spec.minimum;
   }
-  if (spec.exclusive_minimum.has_value()) {
+  // When both bounds are present the larger lower bound wins; on a tie the
+  // exclusive one is stricter.
+  if (spec.exclusive_minimum.has_value() &&
+      (!start.has_value() || *spec.exclusive_minimum >= *start)) {
     start = spec.exclusive_minimum;
+    exclusive_start = true;
   }
   if (spec.maximum.has_value()) {
     end = spec.maximum;
   }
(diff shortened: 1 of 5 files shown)


Ubospica (commented):
Thanks for the fix!

One question I have is, do we need exclusive_start and exclusive_end for float ranges?

Because of rounding error, -1 and -1.00000001 may be the same number, so I feel it's hard to distinguish ranges like [-1, 1] vs (-1, 1). Also, the exclusion depends on the precision of the number, and different precision may lead to different results. I'm wondering your thoughts on this.

Other fixes makes sense to me.

igt0-jota:
The PR https://github.com/mlc-ai/xgrammar/commit/c1bc06c1db00b222125dcf54aa70100a38383066 fixed it, thanks! i am closing this one

Issue #664, opened 2026-06-18
Segfault when TokenizerInfo.serialize_json() and diskcache (stdlib sqlite3) are used in the same process


m-tanner:
## Summary

Constructing an `xgrammar.TokenizerInfo`, calling `.serialize_json()`, and
opening a `diskcache.Cache` (which uses the stdlib `sqlite3` C extension) in the
same process leads to a hard **segmentation fault** (`SIGSEGV`, core dumped) on a
subsequent `TokenizerInfo` construction. Neither library on its own reproduces
it — it only occurs when both xgrammar native code (`TokenizerInfo.serialize_json`)
and `sqlite3`/`diskcache` are exercised in one process. The symptoms (the crash
lands on a *later* allocation, and shrinking the script makes it disappear) are
consistent with **native heap corruption** rather than a deterministic logic bug.

This surfaced while wiring xgrammar's `serialize_json`/`deserialize_json` into a
persistent on-disk grammar cache (the cache backend is `diskcache`).

## Environment

| | |
|---|---|
| xgrammar | 0.2.2 |
| diskcache | 5.6.3 |
| sqlite3 (stdlib) | 3.45.1 |
| transformers | 5.12.1 |
| torch | 2.11.0+cu130 |
| Python | 3.12.3 |
| Platform | Linux x86_64, NVIDIA L4 (crash is on the CPU path; GPU not required — reproduces with `CUDA_VISIBLE_DEVICES=""`) |

## Reproduction (verified — segfaults)

```python
import xgrammar as xgr
from diskcache import Cache
from transformers import AutoTokenizer

tok = AutoTokenizer.from_pretrained("gpt2")

for i in range(2):
    ti = xgr.TokenizerInfo.from_huggingface(tok, vocab_size=50257)
    ti.serialize_json()
    version = xgr.get_serialization_version()
    cache = Cache(
        "/tmp/xgr_diskcache_bug",
        size_limit=512 * 1024 * 1024,
        eviction_policy="least-recently-used",
    )
    if cache.get("__version__") != version:
        cache.clear()
    cache.set("__version__", version)
    print(f"iteration {i} ok", flush=True)

print("survived (no crash)")
```

**Observed:**

```
iteration 0 ok
Segmentation fault (core dumped)
```

The crash occurs during iteration 1 — inside
`xgr.TokenizerInfo.from_huggingface(...)` — i.e. constructing a second
[...]

PR #666, opened 2026-06-18
fix: fix the returned type of serialzation.

Seven-Streams:
As reported in #664, this PR fixes the serialization issue. The root reason is that tvm-ffi returns an `ffi::string`, which will be broken when being pickled. This PR fixes the issue by explicitly converting the return value to str.


Ubospica (approved):
LGTM

PR #668, opened 2026-06-19
feat: support any_order in json_schema format.

Seven-Streams:
This PR adds a new parameter, `any_order,` to determine whether to relax object property ordering. When False (default), properties must appear in their declared order and are fully validated (no duplicate keys, all required keys present). When True,  the properties can be provided in **any order**, while `require`, repeated keys will not be validated.

For example, for the JSONSchema: 

```json
{
    "type": "object",
    "required": ["name", "age"]
}
```

If `any_order` is `False`, then it can only accept `{"name":..., "age":...}`. When `any_order` is `True`, it can accept both `{"name":..., "age":...}` and `{"age":..., "name":...}`. Note that it will under-constrain the output that `{"name":..., "name":...}` and `{"age":..., "age":...}` will be allowed as well.


Seven-Streams:
\review

python/xgrammar/compiler.py
```diff
@@ -181,6 +182,15 @@ def compile_json_schema(
...
             It should be a positive integer.
 
+        any_order : bool, default: False
+            Whether to relax object property ordering. When False (default), properties must appear
+            in their declared order and are fully validated (no duplicate keys, all required keys
+            present). When True, the required properties come first and the non-required ones
+            (optional fields, plus additional/pattern properties) follow, but the order within each
+            group is free; only the count of required properties is checked (duplicate and missing
+            keys are not validated). minProperties/maxProperties still bound the total count.
+            Applies to every object, including nested ones.
```
Ubospica: 1. False: the order is the schema property order
2. True: parameter can take any order, only enforce keys being valid and value schema corresponds to the keys. Num of params is no less than the required properties, and no more than required + optional properties

Issue #671, opened 2026-06-24
[Bug] deepseek_xml structural tag: regex pattern starting with an empty alternative (e.g. ^$|...) produces invalid EBNF (Expect element, but got |)


imReese:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. I have filled all the required information below.
- [x] 4. The bug report is in English.

### Describe the bug

When a JSON schema string property has a `pattern` whose **first regex alternative is empty** — e.g. `"^$|^https://..."`, a common idiom for "empty string OR a URL" fields — compiling it through a **`deepseek_xml` structural tag** fails:

```
grammar_parser.cc:564: EBNF parser error at line 15, column 17: Expect element, but got |
```

The exact same schema compiles fine through the plain `Grammar.from_json_schema` path. The bug only surfaces on the XML tool-calling path.

**Root cause.** The EBNF emitted for the property (via `testing._deepseek_xml_tool_calling_to_ebnf`) is:

```
root_prop_0 ::= | "h" "t" "t" "p" "s" ":" "/" "/" "x" "." "c" "o" "m" "/"
```

The leading `|` (an empty first alternative) is what the parser rejects. Two things combine to produce it:

1. **`RegexConverter` drops the mid-string `^`/`$` anchors** (warning only). So `^$|^https://...` collapses to an **empty first branch** followed by `|` and the second branch. The converter has a special case that inserts an empty string `""` for an empty alternative, but only when it directly precedes `)` (`cpp/regex_converter.cc:329` — `if the previous character is '|'`). It does **not** handle an empty alternative at the **start** of the rule body, so the result begins with a bare `|`.

2. **The XML path emits the regex without any wrapping.** `XMLToolCallingConverter::GenerateString` (`cpp/json_schema_converter_ext.cc:154`) does `return RegexToEBNF(*spec.pattern, false);` as-is. By contrast, the base `JSONSchemaConverter::GenerateString` wraps the regex in quote literals (`"\"" <regex> "\""`), so the rule body never _starts_ with `|` — which is why the `from_json_schema` path avoids the crash. The XML override loses that protection.

**Suggested fix.** Either:
- In `RegexConverter::Convert()`: emit an empty string `""` for an empty leading alternative as well, parallel to the existing `current_[-1] == '|'` special case, so the converted regex never begins with a bare `|` (more general — also covers any unwrapped use of a converted regex); or
[...]

PR #673, opened 2026-06-24
fix: fix the empty string in regex.

Seven-Streams:
This PR fixes the empty string in regex generation as reported in #671.


imReese:
I noticed this was merged without a review. Just out of curiosity about the GitHub merge workflow, how did this get merged directly?

PR #674, opened 2026-06-25
fix: compile SequenceFormat/OrFormat under C++20

tboerstad:
This makes the code compile under C++20. Both structs hold their children by value as a std::vector<Format>, where Format is the variant that also contains these very structs. Under C++20, std::vector requires its element type to be complete at the point it's instantiated. When the constructors were defined inline in the header, they forced vector<Format> to be instantiated while Format (and the structs it references) was still being defined and therefore incomplete, which is ill-formed. Moving the definitions into the .cc delays that instantiation until everything is fully defined, so the build is happy again.

This stays backwards compatible with C++17. Verified by compiling structural_tag.cc under both -std=c++17 and -std=c++20

### DIFF
--- a/cpp/structural_tag.cc
+++ b/cpp/structural_tag.cc
@@ -126,6 +126,12 @@ picojson::value AnyTextFormat::ToJSON() const {
   return picojson::value(std::move(obj));
 }
 
+// These two constructors are defined here rather than inline because instantiating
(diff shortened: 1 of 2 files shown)


tboerstad:
Thanks again for the work on xgrammar, it's been a year since last time around 😅

Ubospica (approved):
LGTM! Thank you so much for your contribution :)

Issue #678, opened 2026-06-27
[Bug] [ROCm] HSA exception after `apply_token_bitmask_inplace_kernel` in vLLM structured output + ngram spec decode


AndreasKaratzas:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. I have filled all the required information below.
- [x] 4. The bug report is in English.

### Describe the bug

I am opening this because vLLM ROCm CI has repeated recent core dumps in the same structured-output case:

```text
tests/entrypoints/llm/test_struct_output_generate.py::test_structured_output[Qwen/Qwen2.5-1.5B-Instruct-xgrammar-auto-speculative_config8]
```

The failure is intermittent. It is not a deterministic Python exception and I cannot make it fail on every local run. The common pattern in the failing CI logs is that vLLM enters inference, Triton JITs xgrammar's `apply_token_bitmask_inplace_kernel`, and ROCm then reports an HSA hardware exception.

### CI failure signature

These are the relevant lines from one failing vLLM ROCm CI log:

```text
entrypoints/llm/test_struct_output_generate.py::test_structured_output[Qwen/Qwen2.5-1.5B-Instruct-xgrammar-auto-speculative_config8]
(EngineCore pid=13931) WARNING ... Triton kernel JIT compilation during inference: apply_token_bitmask_inplace_kernel
GPU coredump: execvp failed: No such file or directory
GPU coredump: handler exited with error (status: 1)
GPU core dump failed
:0:rocdevice.cpp :3586: ... Callback: Queue 0x7fa028600000 aborting with error : HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception. code: 0x1016
WARNING ... [shutdown] MPClient: engine core exited unexpectedly; starting cleanup
```

### Reproduction

Inside the vLLM ROCm CI test image:

```bash
cd /vllm-workspace/tests
export VLLM_WORKER_MULTIPROC_METHOD=spawn
pytest -v -s 'entrypoints/llm/test_struct_output_generate.py::test_structured_output[Qwen/Qwen2.5-1.5B-Instruct-xgrammar-auto-speculative_config8]'
```

The failing CI job ran this as part of the larger `entrypoints/llm` shard:

```bash
cd /vllm-workspace/tests
export VLLM_WORKER_MULTIPROC_METHOD=spawn
pytest -v -s entrypoints/llm \
[...]

PR #679, opened 2026-06-27
[ROCm] Use a 2D launch for the Triton bitmask kernel

AndreasKaratzas:
This PR changes only the ROCm path of `apply_token_bitmask_inplace_triton.py`. CUDA keeps the existing persistent launch. The current ROCm path uses a persistent 1D grid and loops over work with `tl.range(..., NUM_SMS)`. This patch switches ROCm to a regular 2D launch where each Triton program handles one `(vocab block, row)` pair. The motivation is an intermittent ROCm HSA exception we are seeing through vLLM structured-output inference. The failing vLLM logs consistently show this kernel being JITed shortly before the crash:

```text
Triton kernel JIT compilation during inference: apply_token_bitmask_inplace_kernel
HSA_STATUS_ERROR_EXCEPTION ... code: 0x1016
```

Filed an issue regarding this here:
- https://github.com/mlc-ai/xgrammar/issues/678

I could not reduce this to a standalone xgrammar crash. A direct xgrammar stress script using the same Qwen-shaped tensors did not fail locally. So I would not describe this as a proven xgrammar-only bug fix. The reason I think the PR still makes sense is that it removes the ROCm persistent-loop kernel shape from the suspect path, keeps the behavior unchanged, and did not show a performance regression on MI300.

I compared installed `xgrammar==0.2.1` against this patched source module across 480 GPU-timed cases. The benchmark measures only the in-place bitmask application after logits and bitmasks are already on GPU.

| Slice | Cases | Median delta vs installed |
|---|---:|---:|
| All cases | 480 | `-0.0039 ms` |
| Full-row cases | 300 | `-0.0039 ms` |
| Sparse-index cases | 180 | `-0.0039 ms` |
| Qwen vocab `151936` | 120 | `-0.0039 ms` |

On this MI300 setup, I do not see a steady-state performance regression. The patched ROCm path was slightly faster in this sweep, but the main takeaway is that the launch-shape change is behavior-preserving and performance-neutral for the cases tested.

### DIFF
--- a/python/xgrammar/kernels/apply_token_bitmask_inplace_triton.py
+++ b/python/xgrammar/kernels/apply_token_bitmask_inplace_triton.py
@@ -67,7 +67,9 @@ def apply_token_bitmask_inplace_kernel(
         vocab_mask = offsets < vocab_size
         packed_bitmask_mask = bitmask_offsets < bitmask_strides
         packed_bitmask = tl.load(
-            bitmask_ptr + batch_id * bitmask_strides + bitmask_offsets, packed_bitmask_mask
+            bitmask_ptr + batch_id * bitmask_strides + bitmask_offsets,
+            mask=packed_bitmask_mask,
+            other=0,
         )
         bitmask = ((packed_bitmask[:, None] >> (tl.arange(0, 32)[None, :])) & 1) == 0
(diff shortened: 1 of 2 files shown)


AndreasKaratzas:
i m going to be closing this one. There might be an issue with a particular triton version that vllm was running on.

PR #688, opened 2026-07-06
feat: support max_tokens/max_chars in AnyText/AnyTokensFormat via per-state budgets

Ubospica:
### Purpose

Enable a **think budget**: cap how much a model may emit inside an unbounded region of a structural tag (e.g. a `<think>…</think>` reasoning span). This adds:

- `AnyTokensFormat.max_tokens`: bound the number of tokens of a token-level region.
- `AnyTextFormat.max_tokens`: bound the number of **whole tokens** of a byte-level region. Composes with multi-token `excludes` / end markers (e.g. a `</think>` that spans several tokens).
- `AnyTextFormat.max_chars`: bound the number of **Unicode characters** of a byte-level region.

All budgets are opt-in; unset means unbounded (unchanged behavior).

### Design: budgets are per-parser-state counters

The guiding principle is the same one `kRepeat` already uses: **context that affects matching lives in the `ParserState`, not in side channels**, so that the token mask cache contract — masks are a function of the position only; context is encoded by which states are live, with parent-dependent tokens resolved by the uncertain live-check — keeps holding.

- `AnyTokensFormat.max_tokens` is simply lowered to the existing bounded repetition: `any_tokens ::= any_tokens_inner{0, N}`. No parser changes at all.
- `AnyTextFormat.max_tokens` / `max_chars` are carried on the `TagDispatch` grammar expr. The matcher bumps a per-state token counter at each whole-token boundary (reusing `repeat_count`, which trigger-less TagDispatch rules never use), and `AdvanceFsm` counts characters and refuses to consume further bytes once a state's budget is exhausted — a pure function of state + grammar.

Because the counters live in the parser state and the cached adaptive token masks stay **budget-agnostic** (the budget is applied per state at mask-fill time):

- **Rollback / fork / reset** restore the counters automatically with the state history — no parallel bookkeeping.
- **The rule-level (FSM-hash) cross-grammar cache stays enabled** and is shared between bounded and unbounded regions, in both directions, with no staleness by construction.
- **Multiple budgeted regions per grammar** (and re-entry) work, each region instance counts independently.
- **Performance**: while budget remains, mask fill uses exactly the unbounded fast paths (TagDispatch fast-accept, second slicing, speculative calculation). For `max_tokens` there is *no* slow step at all (an exhausted region state is simply skipped; the end tag comes from the already-scanable parent states). For `max_chars` only the last `max-token-char-count` characters of the region take an exact boundary path that starts from the cached mask and only re-simulates the few candidates that may straddle the boundary (see the measurements below).
- **Serialization**: the budget is part of the grammar; serialized compiled grammars keep enforcing it. The serialization version is bumped (`v14` → `v15`) since the `kTagDispatch` expr layout gains two trailing fields.

### Behavior

- A bounded `any_text` (with or without `excludes`) compiles to a `TagDispatch` carrying the budgets; an unbounded, no-excludes `any_text` still lowers to a plain character-class star (no regression).
- When a budget is reached, the region is forced to end: its enclosing end tag becomes the only valid continuation (not EOS).
- The token budget is atomic at token granularity; a token only counts against a region that already existed before that token (a token that *enters* the region, e.g. ends with the begin tag, does not consume budget).
- `max_chars` counts Unicode codepoints (UTF-8 continuation bytes are free).
- Budgets round-trip through the pydantic models, JSON, and the printed EBNF (`TagDispatch(..., max_tokens=N, max_chars=M)`).

### Performance measurements

Benchmarked against pre-PR `main` (each in its own worktree, release build) with the Llama-3.1-8B tokenizer (128k vocab), feeding a ~2000-token / ~8.2k-char reasoning text through a `<think>…</think>` structural tag, plus regression workloads that use no budgets at all. Per-step `fill_next_token_bitmask` latency (us):

| workload | pre-PR mean / p50 / p99 | this PR mean / p50 / p99 |
| --- | --- | --- |
| think region, unbounded | 1.90 / 1.88 / 2.06 | 1.90 / 1.86 / 2.72 |
| builtin `deepseek_r1` + tools (no budget) | 1.93 / 1.88 / 2.09 | 1.92 / 1.86 / 2.11 |
| complex JSON schema (no budget) | 9.86 / 2.62 / 37.8 | 9.54 / 2.65 / 38.1 |

`accept_token` (~0.8 us) and grammar compile times (interleaved A/B, 10 reps: JSON schema 9.19 vs 9.19 ms, `deepseek_r1` 12.86 vs 13.03 ms, think tag 2.54 vs 2.47 ms) show no measurable difference either, i.e. **no regression for unbudgeted grammars**.

For the budgets themselves:

- `max_tokens=N`: in-region steps are identical to unbounded (1.91 vs 1.90 us mean); the forced-end-tag steps after exhaustion are *faster* (~1.5 us, the region state is skipped and only the end tag remains). No compile-time change (mask entries are shared with the unbounded variant).
- `max_chars=M`: in-region steps are identical to unbounded until the remaining budget drops below the max token character count (128 for this vocab, i.e. the last ~84 of 2000 steps here). Those boundary steps take p50 ≈ 11 us / p90 ≈ 38 us, with only the very last steps (< 10 chars remaining) reaching ~0.5–1.7 ms. A one-time per-`CompiledGrammar` char-count table (~1.4 ms for a 128k vocab) is built lazily and shared across matchers/forks.

The boundary path was additionally cross-validated for correctness on the 128k vocab: at sampled steps inside and outside the `max_chars` boundary window, the generated bitmask was compared bit-by-bit against per-token `accept_token` ground truth over the whole vocab. This caught (and the PR fixes) a mask/accept mismatch: with the kRejected mask store type (common at large vocabs), the boundary handler reconstructed the accepted set as all-ones minus rejected/uncertain, wrongly including token ids outside the sorted vocab — e.g. **EOS could leak into the mask near the char budget boundary** (while `accept_token` correctly rejected it). The reconstruction now clears special/stop/out-of-vocab ids, and a >1000-token-vocab regression test forces the non-kAccepted store types and asserts EOS stays masked through and at the boundary.

### Tests

- Converter: TagDispatch construction for bounded regions, `{0, N}` lowering for `any_tokens`, unbounded no-regression, EBNF/pydantic/JSON round-trips, invalid-value rejection.
- Matcher: budget enforcement with a multi-token end tag (forced end tag, not EOS), zero budgets, rollback/fork/reset, bidirectional compiler-cache sharing, multiple bounded regions in one grammar, `max_chars` with multi-char and multi-byte (UTF-8) tokens, hybrid content-plus-end-tag tokens at the boundary, no EOS leak into the boundary mask with a large (>1000-token) vocab, combined `max_tokens`+`max_chars`, compiled-grammar serialization round-trip.
- All existing suites pass (converter, matchers, builtin structural tags, serialization, C++ tests), also with `XGRAMMAR_ENABLE_INTERNAL_CHECK=ON`.

### DIFF
--- a/cpp/grammar_compiler.cc
+++ b/cpp/grammar_compiler.cc
@@ -54,7 +54,14 @@ class GrammarMatcherForTokenMaskCache : public EarleyParser {
         tag_dispatch_rule_id_to_second_slicing_bitset_(tag_dispatch_rule_id_to_second_slicing_bitset
         ),
         tokenizer_info_(tokenizer_info),
-        rule_level_cache_(rule_level_cache) {}
+        rule_level_cache_(rule_level_cache) {
+    // The adaptive token mask cache entries must be budget-agnostic: the cache key ignores the
+    // per-state budget counters, so one entry per position is shared across all counter values
+    // (and across bounded/unbounded rules with the same FSM in the rule-level cache). Budgets are
+    // enforced at runtime by the matcher's parser instead (skip / live-check in
+    // FillNextTokenBitmask, and AdvanceFsm during AcceptToken).
+    SetEnforceBudgets(false);
+  }
   /*!
    * \brief Get the adaptive token mask for the given ParserState.
    * \param is_root_rule Whether to consider the parent rule. If false, there will be
--- a/cpp/compiled_grammar_impl.h
+++ b/cpp/compiled_grammar_impl.h
@@ -10,6 +10,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <mutex>
 #include <string>
 #include <unordered_map>
 #include <utility>
@@ -117,6 +118,20 @@ class CompiledGrammar::Impl {
 
   /*! \brief Mapping from the parser state to the adaptive token mask. */
   std::unordered_map<ParserState, AdaptiveTokenMask, StateHashForCache> adaptive_token_mask_cache;
+
+  /*!
+   * \brief Lazily built per-vocab character (codepoint) count data, used by the matcher for
+   * character budget (TagDispatch::max_chars) boundary handling. Shared by all matchers of this
+   * compiled grammar; built thread-safely on first use; not serialized (rebuilt on demand).
+   * token_char_counts[i] is the character count of sorted vocab token i;
+   * sorted_indices_by_char_count lists all sorted vocab indices ordered by that count, with
+   * char_count_offsets[c] giving the first position whose count is >= c.
+   */
+  std::once_flag token_char_data_once;
+  int32_t max_token_chars = -1;
+  std::vector<int32_t> token_char_counts;
+  std::vector<int32_t> sorted_indices_by_char_count;
+  std::vector<int32_t> char_count_offsets;
 
   Grammar GetGrammar() const { return grammar; }
 
--- a/cpp/earley_parser.cc
+++ b/cpp/earley_parser.cc
@@ -296,6 +296,7 @@ EarleyParser::EarleyParser(
     const Grammar& grammar, const ParserState& init_state, const bool need_expand
 )
     : grammar_(grammar) {
+  has_budgets_ = !grammar_->per_rule_budget_max_tokens.empty();
   if (!grammar->optimized) {
     XGRAMMAR_LOG(FATAL) << "The grammar is not optimized. Please optimize the grammar before using "
                            "the Earley parser.";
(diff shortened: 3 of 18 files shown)


cpp/grammar_compiler.cc
```diff
@@ -54,7 +54,14 @@ class GrammarMatcherForTokenMaskCache : public EarleyParser {
...
         ),
         tokenizer_info_(tokenizer_info),
-        rule_level_cache_(rule_level_cache) {}
+        rule_level_cache_(rule_level_cache) {
+    // The adaptive token mask cache entries must be budget-agnostic: the cache key ignores the
+    // per-state budget counters, so one entry per position is shared across all counter values
+    // (and across bounded/unbounded rules with the same FSM in the rule-level cache). Budgets are
+    // enforced at runtime by the matcher's parser instead (skip / live-check in
+    // FillNextTokenBitmask, and AdvanceFsm during AcceptToken).
+    SetEnforceBudgets(false);
```
Seven-Streams, line 63: `SetEnforceBudgets` seems unecessary, since the function is only called once here.

Seven-Streams (commented):
The logic lgtm. maybe we need some efficiency tests.

Issue #611, opened 2026-04-29
[Bug] GrammarMatcher crashes with invalid token id during JSON mode generation


wopelo:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. I have filled all the required information below.
- [x] 4. The bug report is in English.

### Describe the bug

This issue was originally reported in [mlc-ai/web-llm#807](https://github.com/mlc-ai/web-llm/issues/807), and the web-llm maintainer suggested opening an issue here for better visibility.

### Description
When using JSON mode (response_format: { type: "json_object" }), the GrammarMatcher crashes with an invalid token id. The issue has been reproduced across multiple environments:

- WebLLM (browser, WASM): crashes intermittently
- mlc-llm Python SDK (native, CPU): crashes very frequently — almost every run

The issue is not language-specific (both Chinese and English inputs trigger it) and not quantization-specific (reproduces with both q4f16_1 and q0f16).

### Error Messages
mlc-llm Python SDK (v0.20.dev159):
```
tvm.error.InternalError: xgrammar/cpp/grammar_matcher.cc:245: Check failed:
(token_id >= 0 && token_id < static_cast<int>(compiled_grammar_->vocab_size)) is false:
Invalid token id 151757 for GrammarMatcher
```

WebLLM v0.2.80:
```
[FATAL] xgrammar/cpp/matcher.cc:273: Check failed:
(token_id >= 0 && token_id < tokenizer_info_.GetVocabSize()) is false:
Invalid token id -1447643207 for GrammarMatcherAborted()
```

WebLLM v0.2.82 (latest):
```
[WARNING] xgrammar/cpp/grammar_matcher.cc:499: The token id 1739415078 is out of range [0, 151936). Rejecting the token.
Error: Grammar matcher rejected the newly sampled token.
```

### Steps to Reproduce
Using mlc-llm Python SDK (easiest to reproduce):
```python
from mlc_llm import MLCEngine
engine = MLCEngine("Qwen3-0.6B-q4f16_1-MLC")

response = engine.chat.completions.create(
    messages=[
        {"role": "system", "content": "You are a helpful assistant. Output valid JSON only."},
        {"role": "user", "content": '为名为"发光塑料青蛙玩具，适合6岁以下小朋友，安全无毒"的产品生成营销策略，输出 JSON 包含 title, selling_points, price_range 字段。'}
    ],
[...]

PR #694, opened 2026-07-08
feat: warn when apply_token_bitmask_inplace leaves logits unmasked (undersized bitmask)

whycoming:
## Summary

Addresses the crash class reported in #611 (and upstream mlc-ai/web-llm#807).

`apply_token_bitmask_inplace(logits, bitmask)` auto-detects the vocab size as
`min(logits.shape[-1], bitmask.shape[-1] * 32)` when `vocab_size` is not passed. If the bitmask
is narrower than the logits, the trailing logits `[bitmask.shape[-1] * 32, logits.shape[-1])` are
**silently left unmasked**, so out-of-grammar tokens there (e.g. padded/reserved vocabulary ids)
can still be sampled and are then rejected/aborted by the matcher.

This is exactly what happens in #611: a caller allocates the bitmask with the tokenizer length
(Qwen3 = 151669) while the model emits logits of the padded width (151936). Token id 151757 sits
in the unmasked tail, gets sampled, and triggers `Invalid token id 151757 for GrammarMatcher` /
`Grammar matcher rejected the newly sampled token`.

Minimal repro:

```python
import torch, xgrammar as xgr

logits = torch.ones(1, 151936)                   # model's padded logit width
bitmask = xgr.allocate_token_bitmask(1, 151669)  # sized from tokenizer length (the bug)
bitmask.fill_(0)
xgr.apply_token_bitmask_inplace(logits, bitmask)
assert torch.isfinite(logits[0, 151757])         # reserved id left unmasked -> sampleable
```

## Change

- Emit a `UserWarning` from `apply_token_bitmask_inplace` when `vocab_size is None` **and**
  `bitmask.shape[-1] * 32 < logits.shape[-1]`, pointing the caller to size the bitmask to the
  model's full (padded) vocabulary (or pass the real, unpadded `vocab_size` to intentionally mask
  only a prefix).
- Document the pitfall in the "Padding" note of the docstring.

**No change to masking behavior.** The "logits wider than bitmask" path is intentionally supported
and left untouched; the warning is purely diagnostic and is skipped when `vocab_size` is passed
explicitly. It fires at most once per call site per process (Python's default warning dedup), and
keys only on the vocab dimension, so dynamic batch sizes do not defeat the dedup.

## Tests

Added to `tests/python/test_token_bitmask_operations.py`:

- `test_apply_token_bitmask_inplace_warns_on_undersized_bitmask` and `..._batched` — warns, and
  masking is unchanged (the tail stays finite).
- `test_apply_token_bitmask_inplace_no_warn_when_bitmask_covers_logits` (parametrized over wider /
  exactly-covering) — no warning at or above exact coverage.
- `test_apply_token_bitmask_inplace_no_warn_when_vocab_size_given` — explicit `vocab_size`
  suppresses the warning.

All pass on the CPU backend.

## Open question for maintainers

Since the wider-logits path is a supported feature, I kept this as a **diagnostic warning only**.
If you'd prefer it be doc-only, or gated behind an opt-in strict flag / opt-out kwarg, I'm happy to
adjust.

### DIFF
--- a/python/xgrammar/matcher.py
+++ b/python/xgrammar/matcher.py
@@ -88,7 +88,11 @@ def apply_token_bitmask_inplace(
         will be applied to logits[..., :vocab_size] and bitmask[..., :ceil(vocab_size / 32)].
 
         If vocab_size is not provided, the vocab size will be detected as min(logits.shape[-1],
-        bitmask.shape[-1] * 32).
+        bitmask.shape[-1] * 32). In this case, if the bitmask covers fewer tokens than
+        logits.shape[-1], the trailing logits are left unmasked and out-of-grammar tokens there
+        (e.g. padded/reserved vocabulary ids) can still be sampled, which may cause the sampled
+        token to be rejected downstream. Allocate the bitmask with the model's full (padded)
+        vocabulary size to avoid this. A warning is emitted when this mismatch is detected.
 
     Indices:
         Indices can be used to specify which logits in the batch to apply the bitmask to. It is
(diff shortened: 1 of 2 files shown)


python/xgrammar/matcher.py
```diff
@@ -146,6 +150,20 @@ def apply_token_bitmask_inplace(
             + f"But got logits.device: {logits.device}, bitmask.device: {bitmask.device}"
         )
 
+    if vocab_size is None:
+        masked_width = bitmask.shape[-1] * 32
+        if masked_width < logits.shape[-1]:
+            warnings.warn(
+                f"The bitmask covers fewer tokens (bitmask.shape[-1] * 32 = {masked_width}) than "
```
Seven-Streams: Maybe the warning could be shorter

Seven-Streams:
Thanks for your contribution! I think the PR can be merged after the comment is resolved.

Seven-Streams (approved):

PR #695, opened 2026-07-09
fix(structural_tag): fix several models with `tool_choice="required"` to terminate instead of looping.

Napuh:
With `tool_choice="required"`, some models like GLM models (running on vLLM) that use the GLM-4.7 tool call parser emits the same tool call repeatedly until `max_tokens` is reached (`finish_reason="length"`). `tool_choice="auto"` is unaffected and works fine.

## Root cause

In functions like `get_glm_4_7_structural_tag`, the `required` branch built the suffix with `TagsWithSeparatorFormat(tags=tags, separator="", at_least_one=True)`. That format matches only tool-call tags back-to-back with no free text allowed, so after `</tool_call>` the only grammar-legal continuations are "open another `<tool_call>`" or "end of grammar". There is no grammar-compatible way to emit the end-of-turn token, so it re-opens `<tool_call>` forever.

The `auto` branch already uses `TriggeredTagsFormat`, which allows arbitrary free text between/after tool calls, so the model can terminate naturally. The grammar already permits 1..N calls under `required`, the defect is that it gives the model no possible path to terminate with text.

## Fix

Make `required` use the same `TriggeredTagsFormat` as `auto`, plus `at_least_one=True` (a field `TriggeredTagsFormat` already supports). This preserves the `required` guarantee of >=1 tool call, preserves parallel and chained tool calls, and allows the model to terminate with trailing free text / its natural EOS.

### DIFF
--- a/python/xgrammar/builtin_structural_tag.py
+++ b/python/xgrammar/builtin_structural_tag.py
@@ -625,7 +625,12 @@ def get_llama_structural_tag(
                 )
             )
         assert len(tags) > 0
-        suffix_tag = TagsWithSeparatorFormat(tags=tags, separator="", at_least_one=True)
+        suffix_tag = TriggeredTagsFormat(
+            triggers=[TOOLS_TRIGGER],
+            tags=tags,
+            excludes=_text_excludes(exclude_special_tokens, THINK_EXCLUDE_TOKENS),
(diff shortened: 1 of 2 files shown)


Seven-Streams (approved):
Thanks for your contribution!

PR #697, opened 2026-07-14
Bugfix(fsm): resolve trie back edges via Aho-Corasick failure links

izhuhaoran:
## Bug

`any_text` / `TagDispatch` `excludes` fail to reject a string when the excluded pattern's match starts inside an already-followed trie branch of another pattern.

Minimal repro:

```python
import json, xgrammar as xgr

tag = {"format": {"type": "any_text", "excludes": ["bcd", "abce"]}}
grammar = xgr.Grammar.from_structural_tag(json.dumps(tag))
assert not _is_grammar_accept_string(grammar, "abcd")  # fails: accepted, but contains "bcd"
```

A real-world shape (structural tags for reasoning models): with think-content excludes `["｜DSML｜", "<｜end▁of▁sentence｜>"]`, text containing `<｜DSML｜tool_calls>` is wrongly accepted, so a grammar meant to keep DSML markup out of the think span does not.

## Root cause

`TrieFSMBuilderImpl::AddBackEdges` is not a full Aho-Corasick construction. On a mismatch it only falls back to the start state, plus the start state's depth-1 children (copied onto every state). Any suffix of length >= 2 that is still a prefix of another pattern is lost.

For `excludes=("bcd", "abce")`, input `"abcd"`: the matcher follows the `abce` branch for `abc`, mismatches on `d`, and drops back to the start state — forgetting it just consumed `bc`. The correct failure transition is to the `bc` state of the `bcd` branch, where `d` completes the match.

## Fix

Compute proper failure links and the fully resolved transition table with the textbook `O(num_states * 256)` BFS dynamic program (`delta[u][b] = trie child if present, else delta[fail[u]][b]` — `fail[u]` is strictly shallower, so its row is final when `u` is processed), then materialize each non-end state's row as range edges.

The output is still a deterministic full-coverage FSM of the same shape as before, so match-time behavior and cost are unchanged; for tries without cross-branch suffix overlaps the produced FSM is byte-identical (the pinned FSM-print expectations in `test_fsm_builder.cc` pass unchanged).

## Testing

- New regression test `test_excludes_overlapping_prefixes` (ASCII minimal case + the multi-byte UTF-8 structural-tag case).
- C++ tests: 47/47 pass, including the pinned `TagDispatch` FSM prints.
- Python suite: 2439 passed, 0 failed (HF-gated tests skipped).

### DIFF
--- a/cpp/fsm_builder.cc
+++ b/cpp/fsm_builder.cc
@@ -6,6 +6,7 @@
 
 #include <sys/types.h>
 
+#include <array>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


Seven-Streams (approved):
LGTM! Thanks a lot!

PR #698, opened 2026-07-14
[PoC] add json schema to structural tag to public API

arpera:
Currently public API exposes only `get_model_structural_tag` to process tool calls with reasoning filtering. But there is also an important case where reasoning filtering on Xgrammar's side can help to reduce complexity from vLLM - structured output with json schema set. This PR is a PoC that demonstrates this new feature `json_schema_to_structural_tag`. It converts input json schema to structural tag filtering reasoning. If this idea makes sense and viable then I will complete the implementation of this feature since now it supports only Qwen3.5 model as an example.

How to use `json_schema_to_structural_tag`:
<details>
<summary>main.py</summary>

```python
#!/usr/bin/env python3

from xgrammar import get_model_structural_tag, json_schema_to_structural_tag
from rich.console import Console
import json

tag = json_schema_to_structural_tag(
    model="qwen_3_5",
    schema={
        "type": "object",
        "properties": {"location": {"type": "string"}},
    },
    reasoning=True,
)

data = json.loads(tag.model_dump_json())
Console().print_json(data=data)
```
</details>

**Result:**

<details>
<summary>output</summary>

```json
{
  "type": "structural_tag",
  "format": {
    "type": "sequence",
    "elements": [
      {
        "type": "sequence",
        "elements": [
          {
            "type": "tag",
            "begin": "",
            "content": {
              "type": "any_text",
              "excludes": []
            },
            "end": "</think>"
          },
          {
            "type": "const_string",
            "value": "\n\n"
          }
        ]
      },
      {
        "type": "json_schema",
        "json_schema": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string"
            }
          }
        },
        "style": "json",
        "any_order": false,
        "max_whitespace_cnt": null
      }
    ]
  }
}
```
</details>

### DIFF
--- a/python/xgrammar/json_schema_to_structural_tag.py
+++ b/python/xgrammar/json_schema_to_structural_tag.py
@@ -1,0 +1,53 @@
+from typing import Any, Dict, Optional
+
+from xgrammar.structural_tag import (
+    AnyTextFormat,
+    ConstStringFormat,
+    JSONSchemaFormat,
+    SequenceFormat,
+    StructuralTag,
+    TagFormat,
+)
+
+_THINK_TAG_END = "</think>"
+_THINK_SUFFIX = "\n\n"
+
+# Models that emit optional thinking before plain JSON output.
+_QWEN_STYLE_MODELS = frozenset({"qwen_3_5", "qwen_3_coder", "qwen_3"})
+
+
+def json_schema_to_structural_tag(
+    model: str,
+    schema: Dict[str, Any],
+    reasoning: bool = True,
+    *,
+    any_order: bool = False,
+    max_whitespace_cnt: Optional[int] = None,
+) -> StructuralTag:
+    """Build a structural tag that constrains output to a JSON schema.
+
+    For models with optional thinking (e.g. Qwen3.5), ``reasoning=True`` allows
+    free-form thinking before the JSON payload, matching the prefix used by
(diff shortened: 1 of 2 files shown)


python/xgrammar/json_schema_to_structural_tag.py
```diff
@@ -0,0 +1,53 @@
...
+
+def json_schema_to_structural_tag(
+    model: str,
+    schema: Dict[str, Any],
+    reasoning: bool = True,
+    *,
+    any_order: bool = False,
+    max_whitespace_cnt: Optional[int] = None,
+) -> StructuralTag:
+    """Build a structural tag that constrains output to a JSON schema.
```
yzong-rh, line 27: I think the core logic would work for any `TagFormat` (`JSONSchemaFormat`, `RegexFormat`, etc.) right?

Wouldn't it be more generic to allow any `TagFormat` instead of a `schema`?

Seven-Streams:
@arpera Thanks for your contribution! but I'm thinking about the necessary of adding a new api since we have apis like `from_json_schema` and i think `structural_tag` for json schema is also neat now.

arpera:
The reason I added a new API function in this PoC because I need to filter reasoning on xgrammar's side. As far as I see, `from_json_schema` does not have a flag to filter reasoning. But anyway this PoC is just to show an idea.

PR #704, opened 2026-07-18
prevent XML whitespace state explosion

### LINKED ISSUE #682: [Bug] Segfault on AcceptToken
### Checklist

- [ ] 2. The bug has not been fixed in the latest version.

### Describe the bug

Hi team,

We encountered a segfault in release 0.2.2. We don’t have the full context yet, but it appears to have happened during token acceptance. We’ll continue investigating on our side.

Is this a known issue? Any advice or existing workarounds would be greatly appreciated.

```
[dp28] !!!!!!! Segfault encountered !!!!!!!
[dp28]   File "/project/cpp/earley_parser.cc", line 218, in xgrammar::EarleyParser::Scan(xgrammar::ParserState const&, unsigned char)
[dp28]   File "/project/cpp/earley_parser.cc", line 267, in xgrammar::EarleyParser::Advance(unsigned char, bool)
[dp28]   File "/project/cpp/grammar_matcher.cc", line 671, in xgrammar::GrammarMatcher::Impl::AcceptToken(int, bool)
[dp28]   File "/project/build/tvm_ffi.cc", line 618, in operator()
[dp28]   File "<unknown>", line 0, in PyObject_Call
[dp28]   File "<unknown>", line 0, in PyObject_Vectorcall
[dp28]   File "<unknown>", line 0, in _PyEval_EvalFrameDefault
[dp28]   File "<unknown>", line 0, in PyEval_EvalCode
[dp28]   File "<unknown>", line 0, in PyRun_StringFlags
[dp28]   File "<unknown>", line 0, in PyRun_SimpleStringFlags
[dp28]   File "<unknown>", line 0, in Py_RunMain
[dp28]   File "<unknown>", line 0, in Py_BytesMain
[dp28]   File "<unknown>", line 0, in _start
[dp28]   File "<unknown>", line 0, in 0xffffffffffffffff
```

### Reproduction

Doesn't have a mininal repro yet.

### Environment

vLLM 0.23.0 + xgrammar 0.2.2

We modified the tool parser, using `glm_xml` style:

```
    def _build_structural_tag(self) -> str | None:
        """Build structural_tag for xgrammar guided decoding.

        xgrammar (>=0.2.1) supports ``style="glm_xml"`` on a JSON-schema
        format, which expands to GLM-4.7's
        ``<arg_key>k</arg_key><arg_value>v</arg_value>`` calling convention
        and naturally accepts zero-argument tools (the schema's empty
        ``properties`` produces an empty XML body between
        ``<tool_call>name`` and ``</tool_call>``).
        """
        tags: list[dict[str, Any]] = []
        for tool in self.tools:
            name = tool.function.name
            params = tool.function.parameters or {
                "type": "object",
                "properties": {},
            }

            tags.append(
                {
                    "type": "tag",
                    "begin": f"<tool_call>{name}",
                    "content": {
                        "type": "json_schema",
                        "json_schema": params,
                        "style": "glm_xml",
                    },
                    "end": "</tool_call>",
                }
            )

        if not tags:
            return None

        structural_tag = {
            "type": "structural_tag",
            "format": {
                "type": "triggered_tags",
                "triggers": ["<tool_call>"],
                "tags": tags,
                "at_least_one": False,
                "stop_after_first": False,
            },
        }
        return json.dumps(structural_tag)
```

Ubospica:
## Summary

- remove redundant formatting-whitespace repetitions around unconstrained XML string values
- keep arbitrary whitespace valid as string content while preventing Earley state growth
- replace post-mutation debug-only `Compact2DArray` overflow checks with always-on pre-mutation guards
- validate row-size accumulation and serialized CSR dimensions before narrowing to `int32_t`

## Root cause

XML tool-calling grammars generated unconstrained string values in this form:

```text
whitespace* xml_string whitespace*
```

`xml_string` already accepts whitespace up to the closing tag. A run of `n` spaces therefore produced `n + O(1)` distinct active Earley states, while rollback history retained a quadratic number of states. With GLM-4.7's 512-space token, roughly 128 repeated tokens can push the compact history's cumulative `int32_t` offset past `INT32_MAX`. The next row lookup can then feed an invalid `ParserState` to `EarleyParser::Scan`, matching the stack reported in #682.

The converter now uses the original property schema to identify unconstrained strings through rule aliases and omits the redundant whitespace rules. This preserves the accepted language because the string rule itself still consumes arbitrary whitespace.

The compact container is also hardened so any remaining pathological grammar raises a controlled error before its CSR offsets become unrepresentable in Release builds.

Fixes #682.

## Full issue reproduction

The failure was reproduced with an original `v0.2.2` (`4d145cc1`) Release-style build (`RelWithDebInfo`, `XGRAMMAR_ENABLE_INTERNAL_CHECK=OFF`) and an issue-shaped GLM XML structural tag containing one required unconstrained string parameter. The test repeatedly accepts an ordinary token containing 512 ASCII spaces, which exercises the production `AcceptToken` byte-by-byte path.

The fixed branch was run with the same script and input:

| Metric | Original `v0.2.2` | Fixed `cfbeb433` |
|---|---:|---:|
| Requested input | 130 × 512-space tokens | 130 × 512-space tokens |
| Result | `SIGSEGV` during token 128 | all 130 tokens accepted |
| Exit status | 139 | 0 |
| Wall time | 6:20.52 | 11.49 s |
| `AcceptToken` loop | over 368 s before the crash | 0.045 s to complete |
| Peak RSS (`/usr/bin/time`) | 90.6 GiB | 0.894 GiB |
| Final active parser states | grows linearly with whitespace | 2 |

Under the same measurement method, peak process RSS is about 101× lower (approximately 99.0% reduction), and end-to-end wall time is about 33× lower even though the fixed run accepts more input and completes normally. Runtime RSS sampled after initialization is approximately 0.72 GiB; the 0.894 GiB value above is the comparable whole-process maximum.

Machines with less memory may OOM before the vulnerable build reaches the `int32_t` overflow and segfault. The original production request is no longer available, so this does not prove that its input was the same whitespace sequence; it does reproduce the exact reported `Scan -> Advance -> AcceptToken` stack through the identified mechanism.

## Validation

- exact `v0.2.2` Release-style reproduction: `SIGSEGV`, exit status 139, and stack matching #682
- fixed branch accepts all 130 reproduction tokens with constant parser-state count and exits normally
- pinned pre-commit hooks
- isolated wheel: `tests/python/test_function_calling_converter.py` (134 passed)
- isolated wheel: GLM structural-tag parameterized cases (4 passed)
- Release-style C++ build with `XGRAMMAR_ENABLE_INTERNAL_CHECK=OFF`
- complete C++ test suite (48 passed)
- regression verifies 1,024 spaces keep the latest parser state count constant and the closing tag remains accepted
- regression verifies an unrepresentable `Compact2DArray` size raises a controlled error in Release mode

### DIFF
--- a/cpp/json_schema_converter_ext.cc
+++ b/cpp/json_schema_converter_ext.cc
@@ -235,32 +235,15 @@ std::string XMLToolCallingConverter::FormatPropertyKey(const std::string& key) {
   return JSONSchemaConverter::FormatPropertyKey(key);
 }
 
-std::string XMLToolCallingConverter::FormatPropertyValue(
-    const std::string& value_rule, const SchemaSpecPtr& value_schema
-) const {
-  // xml_string accepts every byte sequence up to the closing tag, including whitespace. Wrapping
-  // it in separate whitespace repetitions preserves the language but creates one Earley state for
-  // every possible split between the repetitions and the string body.
-  if (const auto* string_spec = std::get_if<StringSpec>(&value_schema->spec)) {
-    if (!string_spec->pattern.has_value() && !string_spec->format.has_value() &&
-        string_spec->min_length == 0 && string_spec->max_length == -1) {
-      return value_rule;
-    }
-  }
-  std::string whitespace = GetWhitespacePattern();
-  return whitespace + " " + value_rule + " " + whitespace;
-}
-
 std::string XMLToolCallingConverter::FormatProperty(
-    const std::string& key,
-    const std::string& value_rule,
-    const SchemaSpecPtr& value_schema,
-    const std::string& rule_name,
-    int64_t idx
+    const std::string& key, const std::string& value_rule, const std::string& rule_name, int64_t idx
 ) {
   if (nested_object_level_ <= 1) {
     std::string whitespace = GetWhitespacePattern();
-    std::string formatted_value = FormatPropertyValue(value_rule, value_schema);
+    // xml_string already accepts whitespace. Adding whitespace repetitions around it preserves the
+    // language but creates one Earley state for every possible split with the string body.
+    std::string formatted_value =
+        value_rule == kXMLString ? value_rule : whitespace + " " + value_rule + " " + whitespace;
     if (!xml_wrapper_.value_wrapper_prefix.empty()) {
       return "\"" + xml_wrapper_.key_wrapper_prefix + key + xml_wrapper_.key_wrapper_suffix +
              "\" " + whitespace + " \"" + xml_wrapper_.value_wrapper_prefix + "\" " +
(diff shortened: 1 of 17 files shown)


cpp/support/compact_2d_array.h
Seven-Streams, line 1: if it necessary?

cpp/json_schema_converter_ext.cc
```diff
@@ -235,43 +235,65 @@ std::string XMLToolCallingConverter::FormatPropertyKey(const std::string& key) {
   return JSONSchemaConverter::FormatPropertyKey(key);
 }
 
+std::string XMLToolCallingConverter::FormatPropertyValue(
+    const std::string& value_rule, const SchemaSpecPtr& value_schema
+) const {
+  // xml_string accepts every byte sequence up to the closing tag, including whitespace. Wrapping
```
Seven-Streams: The logic here seems suspicious.

PR #705, opened 2026-07-18
misc:update the version of `transformers`.

Seven-Streams:
This PR enlarges the supported version of `transformers`' to 5.

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -27,7 +27,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install torch transformers datasets tqdm requests
+          # Keep the transformers bound in sync with pyproject.toml.
+          pip install torch "transformers>=4.38.0,!=5.0.*" datasets tqdm requests
 
       - name: Run benchmark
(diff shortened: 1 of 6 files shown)


cjackal:
Fix #708

FrederickGeek8 (changes requested):
Two comments:

**Comment 1:** The version is pinned differently in nearly every changed file here. How about the following version bounds are standardized on?
```
transformers>=4.38.0,<5.13
```

**Comment 2:** Are the changes made to `python/grammar/tokenizer_info.py` and `tests/python/test_tokenizer_info.py` relevant to reverting #689 (i.e. the purpose of this PR)? If the changes are not necessary for that, maybe they should be in a different PR (or commit).

Thanks!

_Hand Crafted, just for you, without an LLM._

Seven-Streams:
@FrederickGeek8 thanks for you suggestions!

Seven-Streams:
@FrederickGeek8 Now the versions are consistent. In previous versions, some tokenizers' vocabulary in `transformers>=5` was misclassified as RAW, and that's the reason for modifying these files.

PR #706, opened 2026-07-18
Cache parser state properties

Ubospica:
## What changed

Cache immutable Earley FSM-state properties lazily, use a dense nullable-rule lookup, and reuse both in character and atomic-token transitions. The cache stays matcher-local; compiled grammars remain immutable.

This commit is based on main at 7a5057ed and changes 2 files, +82/-23.

## Why

Mask generation revisits the same states through shared vocabulary prefixes. These queries previously rescanned edges and nullable-rule lists on every visit.

## Correctness

Release C++: 61 passed. Isolated-wheel targeted Python: 131 passed, with 26 credential-dependent cases deselected. The suite covers nullable and repetition edges, UTF-8, structural tags, token and character budgets, fork, rollback, reset, and transactional failures. After #773 merged, the isolated wheel also passed all 15 max_chars tests. All four benchmark workloads have zero before/after behavior differences.

## Performance

Compile time is in milliseconds and mask time is mean per-step latency in microseconds. Before is this commit's direct parent. A ratio above 1.00x means After is faster than Before.

| Workload | Metric | Before | After | After vs Before |
|---|---|---:|---:|---:|
| Large structural tags | Compile (ms) | 359.646 | 353.207 | 1.02x |
| Large structural tags | Mask (us) | 5.310 | 4.838 | 1.10x |
| Small structural tags | Compile (ms) | 4.897 | 4.617 | 1.06x |
| Small structural tags | Mask (us) | 5.828 | 5.337 | 1.09x |
| Large JSON Schema | Compile (ms) | 93.120 | 45.495 | 2.05x |
| Large JSON Schema | Mask (us) | 19.156 | 15.119 | 1.27x |
| Small JSON Schema | Compile (ms) | 55.686 | 45.460 | 1.22x |
| Small JSON Schema | Mask (us) | 121.696 | 93.191 | 1.31x |

Before/after exact behavior differences in these workloads: 0.

### DIFF
--- a/cpp/earley_parser.cc
+++ b/cpp/earley_parser.cc
@@ -263,12 +263,11 @@ std::pair</* scanable */ bool, /* completable */ bool> EarleyParser::Predict(
   // Check if the rule has a corresponding FSM.
   if (state.rule_id != -1) {
     XGRAMMAR_DCHECK(grammar_->per_rule_fsms[state.rule_id].has_value());
-    // Try to expand the fsm.
-    ExpandNextRuleRefElementOnFSM(state, debug_print);
-    const auto& fsm = grammar_->per_rule_fsms[state.rule_id].value();
-    return std::make_pair(
-        fsm.GetFsm().IsScanableState(state.element_id), fsm.GetFsm().IsEndState(state.element_id)
-    );
+    const uint8_t flags = GetFsmStateFlags(state.rule_id, state.element_id);
+    if (flags & kFsmStateNonTerminal) {
+      ExpandNextRuleRefElementOnFSM(state, debug_print);
+    }
+    return std::make_pair(flags & kFsmStateScanable, flags & kFsmStateEnd);
   }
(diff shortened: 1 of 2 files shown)


Seven-Streams (approved):

Ubospica:
CI note: the Python 3.8 matrix failures occur during test collection in `tests/python/test_max_chars.py:17`, where current `main` uses the Python 3.9-only annotation `list[int]`. This PR changes only `cpp/earley_parser.cc` and `cpp/earley_parser.h`; its C++ build/tests complete before that unrelated collection error, and the Python 3.9-3.14 jobs pass. I am leaving the baseline compatibility fix out of this performance PR to keep its scope isolated.

Ubospica:
The independent baseline fix is already available as #773. It is test-only (`+2/-0`) and its latest aggregated checks are green. After #773 merges, the failed Python 3.8 matrix jobs here can be rerun without adding unrelated files to this PR.

PR #713, opened 2026-07-21
feat: support minimax-m3 style tool call structural tag

izhuhaoran:
## Summary

This PR adds first-class MiniMax M3 structural-tag support, including recursive XML tool arguments and exact matching between runtime-generated opening and closing element names.

MiniMax M3 encodes JSON property names directly as XML element names:

```text
]<]minimax[>[<city>Singapore]<]minimax[>[</city>
```

For dynamic properties, the closing name must exactly repeat the runtime-generated opening name. This equality cannot be expressed by plain CFG productions alone.

## End-to-end example

Given the following tool:

```python
tools = [
    {
        "type": "function",
        "function": {
            "name": "create_order",
            "parameters": {
                "type": "object",
                "properties": {
                    "user_id": {"type": "integer"},
                    "shipping": {
                        "type": "object",
                        "properties": {
                            "city": {"type": "string"},
                            "zip": {"type": "string"},
                        },
                        "required": ["city", "zip"],
                    },
                    "items": {
                        "type": "array",
                        "items": {
                            "type": "object",
                            "properties": {
                                "sku": {"type": "string"},
                                "qty": {"type": "integer"},
                            },
                            "required": ["sku", "qty"],
                        },
                    },
                    "metadata": {
                        "type": "object",
                        "additionalProperties": {"type": "string"},
                    },
                },
                "required": ["user_id", "shipping", "items", "metadata"],
            },
        },
    }
]

structural_tag = xgr.get_model_structural_tag(
    "minimax_m3",
    tools=tools,
    tool_choice="required",
    reasoning_mode="auto",
)

compiled = xgr.GrammarCompiler(tokenizer_info).compile_structural_tag(
    structural_tag
)
matcher = xgr.GrammarMatcher(compiled)
```

The grammar accepts an output such as:

```text
<mm:think>Create the requested order.</mm:think>
]<]minimax[>[<tool_call>
]<]minimax[>[<invoke name="create_order">
]<]minimax[>[<user_id>42]<]minimax[>[</user_id>
]<]minimax[>[<shipping>
]<]minimax[>[<city>Singapore]<]minimax[>[</city>
]<]minimax[>[<zip>018956]<]minimax[>[</zip>
]<]minimax[>[</shipping>
]<]minimax[>[<items>
]<]minimax[>[<item>
]<]minimax[>[<sku>book-001]<]minimax[>[</sku>
]<]minimax[>[<qty>2]<]minimax[>[</qty>
]<]minimax[>[</item>
]<]minimax[>[<item>
]<]minimax[>[<sku>pen-007]<]minimax[>[</sku>
]<]minimax[>[<qty>5]<]minimax[>[</qty>
]<]minimax[>[</item>
]<]minimax[>[</items>
]<]minimax[>[<metadata>
]<]minimax[>[<source>mobile]<]minimax[>[</source>
]<]minimax[>[<campaign>may-launch]<]minimax[>[</campaign>
]<]minimax[>[</metadata>
]<]minimax[>[</invoke>
]<]minimax[>[</tool_call>
```

`user_id`, `shipping`, and `items` are fixed schema properties and compile to ordinary literals.

`source` and `campaign` come from `metadata.additionalProperties`, so their names are generated at runtime. The grammar captures each opening name and requires the corresponding closing name to match byte-for-byte. For example, this is rejected:

```text
]<]minimax[>[<campaign>may-launch]<]minimax[>[</source>
```

## Design

This PR introduces a serializable, grammar-local `DynamicTag(...)` expression:

```text
open_prefix + capture(name_rule) + open_suffix
+ content_rule
+ close_prefix + backreference(captured_name) + close_suffix
```

Each occurrence is lowered into FSM edges:

```text
CaptureStart → name DFA → CaptureEnd → content → BackReference
```

Runtime name equality is therefore part of the grammar instead of being enforced by a grammar-global MiniMax sidecar matcher.

This provides correct occurrence-local behavior for:

- choices/unions and concatenation;
- nested and sibling dynamic elements;
- recursive element content;
- token masks and tokens crossing structural boundaries;
- rollback, fork, reset, jump-forward, batch matching, and EOS;
- grammar optimization, compiler caching, and serialization.

Fixed property names remain ordinary literals and incur no dynamic-name overhead.

## Supported MiniMax M3 behavior

The builtin `minimax_m3` structural tag supports:

- recursive objects and arrays;
- fixed and dynamic properties;
- `additionalProperties`, `patternProperties`, and `propertyNames`;
- strings, numbers, booleans, null, `const`, `enum`, and `$ref`;
- `reasoning_mode="enabled" | "disabled" | "auto"`;
- automatic, required, named, and disabled tool choices;
- one or multiple invocations where allowed.

The implementation reuses `XMLToolCallingConverter` with a recursive M3 dialect rather than introducing a separate schema converter.

Dynamic name rules must be byte-regular so they can be compiled into a bounded DFA and intersected with delimiter-safety constraints. Infinite regular languages such as `[a-z]+` and arbitrary `additionalProperties` keys are supported. Unsupported or unsafe rules fail explicitly during compilation.

## Performance and thread safety

Compiled grammars are immutable and may be shared across threads. Captured names, accepted-byte history, rollback state, and mask scratch storage remain local to each `GrammarMatcher`.

The closing-name mask fast path only checks the vocabulary range matching the next required byte.

Using the local 200,054-token MiniMax tokenizer:

- plain-text accept: `28.83 ns/byte` fixed vs. `31.92 ns/byte` dynamic;
- closing-name mask: approximately `1.96–5.21 μs`;
- dynamic grammar compilation: approximately `50–57 ms`.

`ParserState` size is unchanged.

## Serialization and compatibility

Both `Grammar` and `CompiledGrammar` JSON serialization preserve dynamic-tag semantics. EBNF print/parse also preserves the `DynamicTag(...)` expression.

The serialization version is bumped to `v18`; older serialized grammars must be rebuilt.

## Verification

- C++ tests: `101 passed`
- Full Python suite: `3290 passed, 678 skipped`
- MiniMax M3 focused Python tests: `45 passed`
- Focused ThreadSanitizer tests: `4 passed`, with no race reports
- pre-commit, Ruff, and `git diff --check`: passed

### DIFF
--- a/python/xgrammar/builtin_structural_tag.py
+++ b/python/xgrammar/builtin_structural_tag.py
@@ -15,7 +15,9 @@ from .openai_tool_call_schema import (
 from .structural_tag import (
     AnyTextFormat,
     ConstStringFormat,
+    Format,
     JSONSchemaFormat,
+    OptionalFormat,
     RegexFormat,
     SequenceFormat,
     StructuralTag,
@@ -36,6 +38,7 @@ def get_model_structural_tag(
     any_order: bool = False,
     exclude_special_tokens: bool = True,
     max_whitespace_cnt: Optional[int] = None,
+    reasoning_mode: Optional["ReasoningMode"] = None,
 ) -> StructuralTag:
     r"""Get a structural tag for a model's reasoning and tool-call output format.
 
@@ -181,7 +184,9 @@ def get_model_structural_tag(
         and DeepSeek V4, support both reasoning and non-reasoning modes. If
         ``False``, use the non-reasoning mode. For models that do not support
         reasoning, this has no effect. For models that only support reasoning,
-        ``False`` means reasoning with empty content.
+        ``False`` means reasoning with empty content. For MiniMax M3, this is
+        retained as a compatibility switch: ``False`` selects ``"disabled"``
+        and ``True`` leaves its default ``"auto"`` mode unchanged.
     force_reasoning : bool
         Deprecated. Control whether to keep the reasoning part but leave its content empty.
         Now we will embed the model's specific behavior into the structural tag function, so
@@ -206,12 +211,22 @@ def get_model_structural_tag(
         of whitespace, which avoids the unbounded-whitespace outputs some models
         emit in bad cases that would otherwise blow up grammar
         compilation/matching. Default: ``None``.
+    reasoning_mode : Optional[Literal["enabled", "disabled", "auto"]]
+        MiniMax M3 reasoning mode. ``"enabled"`` requires a reasoning block whose
+        opening marker is already present in the prompt, ``"disabled"`` omits the
+        reasoning block, and ``"auto"`` allows either a complete reasoning block
+        or a direct response/tool call. Defaults to ``None``; MiniMax M3 then uses
+        ``"auto"`` unless ``reasoning=False`` is supplied. Other model formats do
+        not support ``"auto"``; their ``"enabled"``/``"disabled"`` values map to
+        the existing boolean ``reasoning`` behavior.
 
     Notes
     -----
     If a tool's ``parameters`` field is omitted or ``None``, its generated
     arguments are unconstrained JSON. If a function tool has ``strict=False``,
-    its ``parameters`` schema is also treated as unconstrained.
+    its ``parameters`` schema is also treated as unconstrained. MiniMax M3
+    encodes those dynamic objects recursively and validates matching runtime
+    opening/closing element names in ``GrammarMatcher``.
 
     Returns
     -------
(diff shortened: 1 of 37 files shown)


izhuhaoran:
@Ubospica @Seven-Streams Could you please review this PR when you have time ?

Seven-Streams:
@izhuhaoran i'll start reviewing i soon

Seven-Streams:
@izhuhaoran Thanks for your continuous contributions! I think maybe we can break the PR into two parts. The first part is to support minimax-m3 without dynamic tags(for additional properties, we do not require the exact tag). The second part is support for dynamic tags. I think the first part could be merged soon after review. The second part should be further checked.

izhuhaoran:
> @izhuhaoran Thanks for your continuous contributions! I think maybe we can break the PR into two parts. The first part is to support minimax-m3 without dynamic tags(for additional properties, we do not require the exact tag). The second part is support for dynamic tags. I think the first part could be merged soon after review. The second part should be further checked.

@Seven-Streams  Thanks for the suggestion — that makes sense. I’ll split the work into two PRs later

izhuhaoran:
@Seven-Streams Done — I have now split the original implementation as suggested.

This PR contains only the first part:

- MiniMax M3 reasoning and tool-call structural tags
- recursive XML conversion for schemas with fixed property names
- no `DynamicTag` or runtime back-reference machinery

Runtime-named properties such as `additionalProperties` are moved to a stacked follow-up PR, which adds generic `DynamicTag` support and exact opening/closing tag matching. The further PR is https://github.com/mlc-ai/xgrammar/pull/868

Thanks again for the suggestion!

python/xgrammar/builtin_structural_tag.py
```diff
@@ -177,12 +178,13 @@ def get_model_structural_tag(
           available tools before applying its ``mode``. Its ``tools`` list may
           contain both function refs and builtin refs. Builtin refs are matched
           by ``type``.
-    reasoning : bool
-        Whether to enable the reasoning part. Some models, such as Qwen 3.6
-        and DeepSeek V4, support both reasoning and non-reasoning modes. If
-        ``False``, use the non-reasoning mode. For models that do not support
-        reasoning, this has no effect. For models that only support reasoning,
-        ``False`` means reasoning with empty content.
+    reasoning : Union[bool, Literal["enabled", "disabled", "auto"]]
```
Seven-Streams, line 181: In this case, maybe we can mark the strings as the recommended ones, while booleans are `deprecated`?

Seven-Streams (commented):
btw please add some tests in `test_builtin_structural_tag_alignment.py`. It seems that the new feature will break the original function of `kimi-k2` in `test_builtin_structural_tag_alignment.py`.

Seven-Streams (approved):

izhuhaoran:
@Seven-Streams Sorry for missing your comments and being slow to follow up on this PR. Thanks a lot for reviewing it, making the fixes, and getting it merged!

PR #723, opened 2026-07-23
feat: add support for NPU backend in token bitmask application

Yanguan619:
- Introduced a new condition to handle the NPU device type, setting the backend to "torch_native" for improved compatibility.


Seven-Streams:
@Yanguan619 Thanks for your contribution! Have you ever tested in on the npu machines?

Yanguan619:
> @Yanguan619 Thanks for your contribution! Have you ever tested in on the npu machines?

@Seven-Streams Yes, we tested this while using vLLM. When the device is set to NPU, we cannot use the default torch.compile because our device is an NPU, so we use torch_native instead.

Seven-Streams (approved):

PR #733, opened 2026-07-23
feat: Add support for Cohere XML Tool Call Format

jasonozuzu-cohere:
Adds support for Cohere XML Tool Calling Format to the in-built structural tag capabilities in XGrammar. This format differs from other providers in that we rely on nested xml tags to render objects and arrays, which requires subclassing the `XMLToolCallingConverter`. An example tool call is listed below:

```
<|START_THINKING|>...reasoning...<|END_THINKING|>
<cofl:tool_calls>
  <cofl:tool_call id="call-0" name="get_status"></cofl:tool_call>
  <cofl:tool_call id="call-1" name="search">
    <cofl:value name="query" type="raw">nestedxml</cofl:value>
  </cofl:tool_call>
  <cofl:tool_call id="call-2" name="summarize">
    <cofl:value name="documents" type="list">
      <cofl:value type="dict">
        <cofl:value name="id" type="raw">doc-1</cofl:value>
      </cofl:value>
      <cofl:value type="dict">
        <cofl:value name="id" type="raw">doc-2</cofl:value>
      </cofl:value>
    </cofl:value>
    <cofl:value name="options" type="dict">
      <cofl:value name="max_tokens" type="json">200</cofl:value>
    </cofl:value>
  </cofl:tool_call>
</cofl:tool_calls>
```


Seven-Streams:
Thanks for your contribution! I think the supported format is the CMD5 Tool-Calling Format Specification; maybe we should point it out in the doc. btw, I'm not sure if we need to constrain the type part of each parameter. For example, if the parameter is constrained by an `anyof`, which contains different types of values, the current constraint seems not accurate.

jasonozuzu-cohere:
@Seven-Streams thanks for the feedback! 🙏🏾

I don't think we want to publically document the CMD5 moniker, as this is something we only use internally - perhaps this is something to update when we finalise the name for the model/model family.

Also, have refactored to flesh out support for `anyOf` as well as some of the other ambiguous json schema types. The `type` field is important to us as it aids in parsing, so definitely something we want to ensure is correct relative to the value in the xml tag.

Seven-Streams:
@jasonozuzu-cohere Thanks a lot! I think most of it is okay now. But it seems that `patternproperties` cannot be supported properly yet(in the `<cofl:value name=` part). I think maybe we can either add a warning about it(since it may not appear in tool-calling as the constraint for the parameter name?), or try to figure out how to fix it.

jasonozuzu-cohere:
@Seven-Streams Thanks, good catch! Took a look at the implementation and managed to refactor to support `patternProperties` and `propertyNames`. Is there anything else we need to cover?

Seven-Streams (approved):

PR #750, opened 2026-07-28
Support Unicode patterns in RegexFSMBuilder

Ubospica:
## Summary
- parse `RegexFSMBuilder` patterns as Unicode code points and encode scalar values and ranges as valid UTF-8 byte paths
- support Unicode literals, positive and negative character classes, `.`, and existing Unicode escape forms while rejecting invalid UTF-8, surrogate values, and out-of-range escapes
- allow Unicode JSON Schema patterns to use `Regex(..., json_string=true)` while retaining the existing grammar-expansion fallback for unsupported patterns

## Test plan
- [x] `cmake --build build-tests -j 16`
- [x] `ctest --test-dir build-tests --output-on-failure` (66 passed)
- [x] `python -m pytest tests/python/test_grammar_regex_macro.py tests/python/test_regex_converter.py tests/python/test_grammar_matcher_regex.py -m \"not hf_token_required\" -q` (210 passed, 16 deselected)

### DIFF
--- a/cpp/fsm_builder.cc
+++ b/cpp/fsm_builder.cc
@@ -6,23 +6,159 @@
 
 #include <sys/types.h>
 
+#include <algorithm>
 #include <array>
 #include <cstddef>
 #include <cstdint>
-#include <cstring>
+#include <initializer_list>
+#include <limits>
 #include <optional>
 #include <set>
 #include <stack>
+#include <unordered_map>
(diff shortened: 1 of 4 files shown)


Ubospica:
cc @Seven-Streams this PR can deprecate the legacy regex path. Please check it out.

PR #764, opened 2026-07-30
Support regex intersections and complements in Lark and EBNF

Ubospica:
## Summary

- support `&` intersections and prefix `~` complements of regular expressions in both Lark and
  EBNF grammars
- represent them as native `kIntersect` / `kComplement` grammar exprs; operands compile through
  the codepoint-aware regex pipeline into leaf automata, which are intersected or negated
  directly (complements are taken with respect to all valid UTF-8 strings)
- run the automaton intersection directly on NFAs instead of determinizing both operands first,
  and fix two latent state-merging bugs in `MergeEquivalentStates` that epsilon NFAs exposed
- enforce automaton state limits and document precedence, scope, and compilation cost for both
  frontends

## Behavior

`&` binds more tightly than `|` and looser than a sequence; prefix `~` applies to one element,
and a quantifier after `~x` applies to the complemented element. Both operators are accepted in
uppercase terminal definitions and directly in rules, where they behave like terminals
(lexemes with a trailing skip). Operands must stay regular: rule references (and, in EBNF,
repetition ranges inside operands) raise source-located errors.

The implementation preserves valid UTF-8 character semantics, including grouped and bounded
regular expressions, Unicode character classes, empty languages, and nullable operands.
Generated grammars round-trip through both JSON serialization and printed EBNF; the
serialization version is unchanged.

## Validation

- C++ test suite: 67 passed, including an exhaustive NFA-product equivalence test (15 automata
  pairwise intersected, checked against both operands on every string up to length 6 over a
  3-letter alphabet) and regression tests for the state-merging fixes
- Python suite excluding environment-dependent tests: 3481 passed, 1 skipped, including
  exhaustive parity tests against Python's `re` module for `A & B`, `~A`, and `A & ~B`
- language parity vs llguidance: intersections (12 grammars / 75 candidates) and complements
  (8 grammars / 53 candidates) matched

### DIFF
--- a/cpp/fsm.cc
+++ b/cpp/fsm.cc
@@ -1154,59 +1154,89 @@ Result<FSMWithStartEnd> FSMWithStartEnd::Intersect(
   if (!lhs.IsLeaf() || !rhs.IsLeaf()) {
     return ResultErr("Intersect only support leaf fsm!");
   }
-  auto lhs_dfa_raw = lhs.ToDFA(max_result_num_states);
-  auto rhs_dfa_raw = rhs.ToDFA(max_result_num_states);
-
-  if (lhs_dfa_raw.IsErr()) {
-    return lhs_dfa_raw;
-  }
-  if (rhs_dfa_raw.IsErr()) {
-    return rhs_dfa_raw;
-  }
-
-  auto lhs_dfa = std::move(lhs_dfa_raw).Unwrap();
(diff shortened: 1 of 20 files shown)


Seven-Streams (commented):
I think the main problem is that `~ ["]` can accept a string with any length? or it is just the semantics.

PR #774, opened 2026-07-30
Reject unresolved JSON Schema references

Ubospica:
## Summary

- reject external document references and unresolved named anchors instead of emitting unconstrained grammars
- treat an empty reference URI as a root self-reference
- preserve existing local JSON Pointer and recursive-reference support

## Root cause

Non-local reference URIs fell back to an unconstrained schema after logging a warning. Invalid input could therefore silently widen accepted output.

## Validation

- reference parity: four unsupported URI forms rejected; empty URI resolves to root
- Python core suite: 3064 passed, 1 skipped, 622 deselected
- JSON Schema converter suite: 523 passed
- serialization suite: 13 passed, 1 deselected
- C++ Debug/InternalCheck: 61 passed
- pre-commit: passed

Depends on #773 for the shared Python 3.8 test compatibility commit; that commit will leave this diff when #773 lands.

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -1439,6 +1439,9 @@ Result<RefSpec, SchemaError> SchemaParser::ParseRef(const picojson::object& sche
   }
   RefSpec spec;
   spec.uri = schema.at("$ref").get<std::string>();
+  if (spec.uri.empty()) {
+    spec.uri = "#";
+  }
   return ResultOk(std::move(spec));
 }
 
@@ -1458,8 +1461,10 @@ Result<SchemaSpecPtr, SchemaError> SchemaParser::ResolveRef(
   }
 
   if (uri.size() < 2 || uri[0] != '#' || uri[1] != '/') {
-    XGRAMMAR_LOG(WARNING) << "URI should either be '#' or start with '#/' but got " << uri;
-    return ResultOk(SchemaSpec::Make(AnySpec{}, "", "any"));
+    return ResultErr<SchemaError>(
+        SchemaErrorType::kInvalidSchema,
+        "Unsupported $ref '" + uri + "': only local references '#' and '#/...' are supported"
+    );
   }
 
   std::vector<std::string> parts;
--- a/tests/python/test_json_schema_converter.py
+++ b/tests/python/test_json_schema_converter.py
@@ -734,6 +734,37 @@ def test_reference_schema():
     )
 
 
+@pytest.mark.parametrize(
+    "uri",
+    ["https://example.com/schema.json", "other.json", "other.json#/$defs/value", "#named-anchor"],
+)
+def test_external_reference_is_rejected(uri):
+    with pytest.raises(RuntimeError) as error:
+        xgr.Grammar.from_json_schema(json.dumps({"$ref": uri}), any_whitespace=False)
+
+    message = str(error.value)
+    assert "only local references '#' and '#/...' are supported" in message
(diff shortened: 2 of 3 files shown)


percymcn:
This still reproduces on **0.2.5** (the current release), so flagging it in case the close was procedural rather than a decision — the diff here looks right to me, and the root-cause line in the description ("non-local reference URIs fell back to an unconstrained schema after logging a warning; invalid input could therefore silently widen accepted output") is exactly what I measured.

Everything below is executed against `xgrammar==0.2.5` under Python 3.10; identical on 0.2.3 and 0.2.4.

**1. The reachable spelling isn't in the test list.** The four parametrised URIs are all recognisably foreign (`https://…`, `other.json`, `other.json#/$defs/value`, `#named-anchor`), which reads as "someone typed a URL". The form that actually shows up in generated schemas is `/$defs/Name` — Pydantic emits it verbatim when given `ref_template="/$defs/{model}"`, and `litellm` 1.96.0 passes exactly that on its Anthropic path (`llms/anthropic/chat/transformation.py`), so any user routing a Pydantic `response_format` through that path and then serving it here gets this shape. It behaves the same as the other four:

```
$ref value                       result on 0.2.5
https://example.com/schema.json  compiles, accepts a number where a string enum was declared
other.json                       compiles, accepts a number
other.json#/$defs/value          compiles, accepts a number
#named-anchor                    compiles, accepts a number
/$defs/P                         compiles, accepts a number     <-- not covered
```

Schema used: `{"type":"object","properties":{"p":{"$ref": URI}},"required":["p"],"$defs":{"P":{"type":"string","enum":["low"]}}}`. Note the ref is at a *property* position rather than the root; the four tests here all use a root `{"$ref": uri}`, and both positions behave identically, but property position is where generators put them.

**2. A test-design note.** Because all four fixtures are "obviously foreign" URIs, a regression suite built on them passes while the one form real generators emit still ships. Adding `/$defs/Value` to the `parametrize` list costs one line and covers the population the PR is actually protecting.

**3. The asymmetry is what makes this hard to notice.** A missing *target* is loud, a malformed *prefix* is silent, for the same absent definition:

```
{"$ref": "#/$defs/Missing"}   RuntimeError: Cannot find field Missing in #/$defs/Missing
{"$ref": "/$defs/Missing"}    compiles; the field accepts any JSON
```

So a user who typos the definition name gets a clear error, and a user whose generator emits the wrong prefix gets a grammar that constrains nothing. Only the stderr `Warning: URI should either be '#' or start with '#/'` distinguishes them, and that is easy to lose in a server log.

**4. The widening isn't confined to the node with the bad `$ref`.** Tooling that merges `$ref` siblings (draft 2020-12 applies the referent *and* the siblings) resolves the pointer with the same "is this a local pointer?" test, so an unrecognised prefix makes the whole node invisible to that merge and the siblings never get folded in. Concretely, `{"$ref": "/$defs/P", "minLength": 5}` reaches the converter unmerged, and:

```
{"p": "ab"}   accepted   (minLength: 5 not enforced)
```

whereas the same document with `#/$defs/P` enforces it. So one malformed prefix can drop constraints that were never written inside the referenced definition at all.

For what it's worth, `outlines-core` 0.2.14 and `lm-format-enforcer` 0.11.3 both resolve `/$defs/P` correctly and both raise on a missing target in either spelling — so a schema that silently loses its constraints here keeps them on those two, which makes the difference easy to miss when a schema is shared across backends.

Happy to open a separate issue instead if that's easier to track, but since the fix is already written here it seemed more useful to attach the measurements to it.

Issue #749, opened 2026-07-28
pattern/format combined with minLength/maxLength silently drops the length bounds


he-yufeng:
## Summary

A string schema that mixes a generative constraint (`pattern` or `format`) with `minLength`/`maxLength` is accepted by `Grammar.from_json_schema`, but the length bounds are silently dropped from the compiled grammar. The model can then return strings that violate the bound with no error anywhere.

## Reproduction

```python
import xgrammar as xgr

g1 = xgr.Grammar.from_json_schema({"type": "string", "pattern": "^a+$", "maxLength": 2})
g2 = xgr.Grammar.from_json_schema({"type": "string", "pattern": "^a+$"})
print(str(g1) == str(g2))  # True: maxLength had no effect
```

Compiled EBNF for both:

```
root_1 ::= (("a" root_1) | ("a"))
```

The `maxLength: 2` bound does not appear anywhere in the grammar. The same holds for `format` combined with length bounds.

## Expected behavior

Either the two constraints compose (the grammar enforces both the pattern and the length bound), or the combination is rejected loudly at compile time. Silently dropping one of them is the worst of the three: the schema says one thing, the grammar does another, and nothing tells the caller.

## Context

Found while validating structured-output requests server side. A request carrying such a schema is accepted today and can produce out-of-bound strings with no signal. Downstream (vLLM) is adding a validation-time guard that rejects the combination with a clear 400 until the composition works here (vllm-project/vllm#49996), but the right home for the real fix is the xgrammar compiler.

maxwillzq:
I'm working on this issue! I've opened PR #798 to reject combining pattern/format with minLength/maxLength at compile-time with a clear error, preventing silent constraint dropping.

Seven-Streams:
Currently Add warnings in #896 and will be closed for now. Feel free to reopen it!

PR #796, opened 2026-08-01
fix: reject pattern/format combined with minLength/maxLength

feiiiiii5:
## Root Cause

`GenerateString` in `cpp/json_schema_converter.cc` picks exactly one branch for string constraints: a built-in `format` or a `pattern` (compiled as a regex) takes priority, and the `minLength`/`maxLength` branch is never reached. A schema like `{"type": "string", "pattern": "^a+$", "maxLength": 2}` was accepted by `Grammar.from_json_schema`, but the compiled grammar ignored `maxLength` entirely — the model could produce out-of-bound strings with no error anywhere (issue #749).

## Fix

Reject the combination loudly at schema parse time instead of silently dropping the bounds (composing regex + length counting is not supported by the grammar engine yet — the same direction vLLM takes downstream with a validation-time guard in vllm-project/vllm#49996):

- `ParseString` now returns `kUnsupportedSchema` when `pattern`, or a **built-in** `format` (one that maps to a regex in `GenerateString`), is combined with non-default `minLength`/`maxLength`.
- New public static `JSONSchemaConverter::IsBuiltinFormat()` exposes the built-in check to the parser. Unknown formats are annotations per JSON Schema and are unaffected — they still fall through to the length branch.

```cpp
const bool has_builtin_format =
    spec.format.has_value() && JSONSchemaConverter::IsBuiltinFormat(*spec.format);
if ((spec.pattern.has_value() || has_builtin_format) &&
    (spec.min_length != 0 || spec.max_length != -1)) {
  return ResultErr<SchemaError>(
      SchemaErrorType::kUnsupportedSchema,
      "combining pattern/format with minLength/maxLength is not supported yet; "
      "the length bounds would be silently ignored. ...");
}
```

## Test

- `test_pattern_with_length_bounds_is_rejected` — pattern + `maxLength` / + `minLength` raise with a clear message.
- `test_builtin_format_with_length_bounds_is_rejected` — built-in `format: "email"` + `maxLength` raises.
- `test_pattern_format_and_length_bounds_alone_still_supported` — pattern alone, format alone, min/max alone, and **unknown** format + bounds still compile (no behavior change outside the combination).
- Full run: `test_json_schema_converter.py` + `test_json_schema_direct_converter.py` → **547 passed**.
- `clang-format` (19.1.7) clean on the changed C++ files; `ruff check` clean.

## Diff scope

3 files, +65/-0 (`cpp/json_schema_converter.cc`, `cpp/json_schema_converter.h`, `tests/python/test_json_schema_converter.py`).

## AI Disclosure

AI-assisted root-cause analysis, initial draft, and test scaffolding; human review of the diff, error-path semantics, and test scope.

## Not PR-related failures

None observed locally; the two converter test files pass in full.

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -1136,6 +1136,22 @@ Result<StringSpec, SchemaError> SchemaParser::ParseString(const picojson::object
             std::to_string(spec.max_length)
     );
   }
+  // GenerateString chooses exactly one branch for string constraints: a built-in
+  // format or a pattern (compiled as a regex) takes priority, and length bounds
+  // are silently dropped. Composing the two constraints (regex + length counting)
+  // is not supported by the grammar engine yet, so fail loudly at parse time
+  // instead of accepting a schema whose length bounds would never be enforced.
+  const bool has_builtin_format =
+      spec.format.has_value() && JSONSchemaConverter::IsBuiltinFormat(*spec.format);
+  if ((spec.pattern.has_value() || has_builtin_format) &&
+      (spec.min_length != 0 || spec.max_length != -1)) {
+    return ResultErr<SchemaError>(
+        SchemaErrorType::kUnsupportedSchema,
+        "combining pattern/format with minLength/maxLength is not supported yet; "
+        "the length bounds would be silently ignored. Remove the pattern/format "
+        "constraint or the length bounds, or use minLength/maxLength alone"
+    );
+  }
   return ResultOk(std::move(spec));
 }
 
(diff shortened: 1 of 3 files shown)


feiiiiii5:
Friendly ping — this PR (reject `pattern`/`format` combined with `minLength`/`maxLength`) is mergeable. Would appreciate a review, or a note if it's not a priority — happy to adjust or close it to reduce noise.

feiiiiii5:
Closing this PR after 14+ days with no human review or comment on the PR, per my contribution hygiene rules (leave the branch, archive the finding). Fix and regression tests are in the PR body. The fix, regression tests, and branch `fix/reject-pattern-format-length-bounds` remain available; happy to reopen and continue if a maintainer picks it up.

PR #798, opened 2026-08-01
fix(json_schema): reject combining pattern/format with length constraints

### LINKED ISSUE #749: pattern/format combined with minLength/maxLength silently drops the length bounds

maxwillzq:
Fixes #749.

### Summary
When a string schema in JSON Schema combines a generative constraint (`pattern` or `format`) with length bounds (`minLength` or `maxLength`), `JSONSchemaConverter::GenerateString` previously returned early for the pattern/format rule and silently dropped the length bounds.

### Fix
This PR updates `SchemaParser::ParseString` in `cpp/json_schema_converter.cc` to return a `SchemaErrorType::kUnsupportedSchema` error when both generative constraints and length bounds are specified. This causes `Grammar.from_json_schema` to raise a clear `RuntimeError` at schema compilation time instead of silently ignoring the length bounds.

### Verification
- Added `test_pattern_format_with_length_constraints_raises` in `tests/python/test_json_schema_converter.py`.
- Ran unit tests: `1 passed`.

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -1111,24 +1111,49 @@ Result<NumberSpec, SchemaError> SchemaParser::ParseNumber(const picojson::object
 
 Result<StringSpec, SchemaError> SchemaParser::ParseString(const picojson::object& schema) {
   StringSpec spec;
+  bool has_generative = schema.count("format") || schema.count("pattern");
+  bool has_length = schema.count("minLength") || schema.count("maxLength");
+
+  if (has_generative && has_length) {
+    return ResultErr<SchemaError>(
+        SchemaErrorType::kUnsupportedSchema,
+        "Combining pattern/format with minLength/maxLength is currently unsupported"
+    );
+  }
+
   if (schema.count("format")) spec.format = schema.at("format").get<std::string>();
   if (schema.count("pattern")) spec.pattern = schema.at("pattern").get<std::string>();
+
   if (schema.count("minLength")) {
     if (!schema.at("minLength").is<int64_t>()) {
       return ResultErr<SchemaError>(
           SchemaErrorType::kInvalidSchema, "minLength must be an integer"
       );
     }
-    spec.min_length = static_cast<int>(schema.at("minLength").get<int64_t>());
+    int64_t min_len = schema.at("minLength").get<int64_t>();
+    if (min_len < 0) {
+      return ResultErr<SchemaError>(
+          SchemaErrorType::kInvalidSchema, "minLength must be a non-negative integer"
+      );
+    }
+    spec.min_length = static_cast<int>(min_len);
   }
+
   if (schema.count("maxLength")) {
     if (!schema.at("maxLength").is<int64_t>()) {
       return ResultErr<SchemaError>(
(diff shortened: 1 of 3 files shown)


maxwillzq:
Updated the PR to address Copilot's review:
1. Added non-negative validation for `minLength` and `maxLength` (returning `kInvalidSchema` if negative).
2. Keyed the unsupported combination check directly on keyword presence (`schema.count(minLength) || schema.count(maxLength)`) rather than sentinel default values.
3. Updated python unit tests accordingly.

maxwillzq:
cc @he-yufeng @Ubospica - Would appreciate your feedback on this fix for #749 when you have a chance!

Seven-Streams (approved):

Seven-Streams:
Wait a minute, we may need further discussion, since in some cases, we may support it, and we do not want to break the behavior.

maxwillzq:
Thanks a lot!

maxwillzq:
any update ? @Seven-Streams  Thanks

maxwillzq:
Hi @Seven-Streams, thanks for the feedback! 

To address your concerns regarding backward compatibility and avoiding breaking valid user schemas:

1. **Permit Trivial Constraints**: We updated the logic so trivial constraints like `minLength: 0` (commonly emitted by Pydantic/OpenAPI) no longer trigger an error and compile cleanly alongside `pattern` / `format`.
2. **Targeted Rejection**: We now only reject combinations where active generative rules (`pattern` or built-in `format`) coexist with *non-trivial* bounds (`minLength > 0` or `maxLength != -1`), which are the exact cases where length bounds are silently dropped (#749). Unknown/custom formats fall through and continue to enforce length bounds as expected.
3. **Safety & Bounds Check**: Added string type validation and `INT_MAX` overflow guards.
4. **Rebased & Conflict Resolved**: Rebased onto the latest `main` with new regression tests added (all 523 converter tests pass).

Could you please take another look when you have a moment? Thanks!

PR #803, opened 2026-08-04
fix(json_schema): emit nothing for property-less XML tool-call objects

### LINKED ISSUE #802: [Bug] XML tool calling loops forever on no-argument tools
### Checklist

### Describe the bug

The XML tool-calling styles put each property in its own `<parameter=key>value</parameter>` tag, so the root object has no braces. When the object can't hold any property, everything drops out and only whitespace is left:

```
root ::= [ \n\t]*
```

That's a self-loop with no non-whitespace way forward. In a structural tag, once generation hits the parameter zone every allowed token is whitespace and none of them move the automaton along. `\n` is in there and would fall through to the tag's `end`, so the grammar isn't blocking the right token — it just has to out-compete `\t` every step, and Qwen usually picks `\t` (in training data `<function=name>\n` is nearly always followed by `\t<parameter=...>`). Once that starts it never recovers and the request burns its whole budget on tabs. Showing up as vllm-project/vllm#50989.

Tools with parameters are fine, they still have `<parameter=` to go to.

The whitespace doesn't mean anything anyway — the template already puts `">\n"` and `"\n</function>"` around the parameter zone.

### Reproduction

No model or tokenizer needed:

```python
from xgrammar.testing import _json_schema_to_ebnf

print(_json_schema_to_ebnf(
    {"type": "object", "properties": {}, "required": []}, json_format="qwen_xml"))
# root ::= [ \n\t]*
# expected: nothing, so the tag's `end` takes over
```

Same output on 0.2.3 under the older helper name `_qwen_xml_tool_calling_to_ebnf(schema)`,
from before the `json_format` argument existed.

Through a structural tag, the parameter zone eats unbounded whitespace:

```python
import xgrammar as xgr
from xgrammar.structural_tag import (
    StructuralTag, TagFormat, TriggeredTagsFormat, JSONSchemaFormat,
)

vocab = [bytes([b]).decode("latin-1") for b in range(256)]
compiler = xgr.GrammarCompiler(xgr.TokenizerInfo(vocab, vocab_type=xgr.VocabType.RAW))
tag = StructuralTag(format=TriggeredTagsFormat(
    triggers=["<tool_call>"],
    tags=[TagFormat(
        begin="<tool_call>\n<function=noarg>\n",
        content=JSONSchemaFormat(
            json_schema={"type": "object", "properties": {}, "required": []},
            style="qwen_xml"),
        end="\n</function>\n</tool_call>")]))

m = xgr.GrammarMatcher(compiler.compile_structural_tag(tag))
for ch in "<tool_call>\n<function=noarg>\n":
    assert m.accept_string(ch)
print(m.accept_string("\t"))  # True, and stays True forever
```

### Environment

- XGrammar 0.2.3 (pinned by vLLM 0.26.0), also on current `main`
- vLLM 0.26.0, `--tool-call-parser qwen3_engine`, `Qwen3.6-35B-A3B-FP8`, Linux / CUDA
- Triggered by a tool with `"strict": true` and `parameters={"type": "object", "properties": {}, "required": []}`

wangxuw:
The XML tool-calling styles emit each property as its own `<parameter=key>value</parameter>` tag, so the root object carries no braces. When the object cannot hold any property, the generic object path collapses to `whitespace*` -- a self-loop with no non-whitespace successor. Inside a structural tag every allowed token is then whitespace and none advances the automaton, so a model whose prior favours indentation keeps emitting tabs until it runs out of budget instead of closing the tool call. `\n` is reachable and would fall through to the tag's end pattern, but it has to out-compete `\t` at every step, which makes the stall intermittent rather than guaranteed. Reported downstream as vllm-project/vllm#50989.

Return `Empty()` for a root object that cannot hold a property, letting the enclosing tag's end pattern take over. The minimal accepted output does not change: the Qwen template already supplies `">\n"` and `"\n</function>"` around the parameter zone, so the parameter zone never needed whitespace of its own.

The guard stays narrow on purpose. `parameters: None` and non-strict function tools map to a `True` schema, and open objects (bare `{"type": "object"}`, `additionalProperties: true` / `{...}`) can genuinely carry properties, so neither is rewritten -- both keep a `<parameter=` branch and therefore a non-whitespace way forward. `min_properties` needs no check because a positive lower bound with no permitted property is already rejected as unsatisfiable. The `json` style is untouched and still requires `{}`.

Tests cover all four XML styles, `$ref` indirection, the `json` style as a control, open objects, nested objects, and an end-to-end structural tag asserting that the parameter zone rejects whitespace while the closing sequence is still accepted.

Fixes #802.

### DIFF
--- a/cpp/json_schema_converter_ext.cc
+++ b/cpp/json_schema_converter_ext.cc
@@ -398,7 +398,22 @@ int32_t XMLToolCallingConverter::GenerateObject(
 ) {
   nested_object_level_++;
   bool need_brace = nested_object_level_ > 1;
-  auto result = JSONSchemaConverter::GenerateObject(spec, rule_name, need_brace);
+  // The root XML object is not wrapped in braces: its properties are emitted as
+  // separate `<parameter=key>value</parameter>` tags. When such an object cannot hold
+  // any property, the generic path degenerates into `whitespace*`, a self-loop with no
+  // non-whitespace successor. A model whose prior favours indentation can then emit
+  // whitespace until its token budget runs out instead of closing the tool call.
+  // Emitting nothing hands control straight to the enclosing structural tag's end
+  // pattern. `min_properties` needs no check here: a positive lower bound with no
+  // permitted property is already rejected as an unsatisfiable schema.
+  bool cannot_contain_properties =
+      !need_brace &&
+      (spec.max_properties == 0 ||
+       (spec.properties.empty() && spec.pattern_properties.empty() && !spec.property_names &&
+        !spec.allow_additional_properties && !spec.allow_unevaluated_properties));
+  auto result = cannot_contain_properties
+                    ? Empty()
+                    : JSONSchemaConverter::GenerateObject(spec, rule_name, need_brace);
   nested_object_level_--;
   return result;
 }
--- a/tests/python/test_empty_xml_object.py
+++ b/tests/python/test_empty_xml_object.py
@@ -1,0 +1,204 @@
+"""Tests for XML tool-calling objects that cannot contain any property.
+
+The XML tool-calling formats compile such an object into an empty string rather
+than ``whitespace*``. A whitespace-only body is a self-loop with no
+non-whitespace successor, so a model whose prior favours indentation can keep
+emitting whitespace until its token budget is exhausted instead of closing the
+tool call.
+"""
+
+import sys
+
+import pytest
+
+import xgrammar as xgr
+from xgrammar.structural_tag import JSONSchemaFormat, StructuralTag, TagFormat, TriggeredTagsFormat
+from xgrammar.testing import _is_grammar_accept_string, _json_schema_to_ebnf
+
+XML_FORMATS = ["qwen_xml", "minimax_xml", "deepseek_xml", "glm_xml", "kimi_k3_xml"]
+
+WHITESPACE_INSTANCES = [" ", "\t", "\n", "\t\t\t", " \n\t"]
+
+# The schema OpenAI-style clients emit for a tool that takes no argument. Note
+# the absence of ``additionalProperties``, which is the shape that surfaced the
+# whitespace loop in practice.
+NO_ARGUMENT_SCHEMA = {"type": "object", "properties": {}, "required": []}
+
+CLOSED_NO_ARGUMENT_SCHEMA = {**NO_ARGUMENT_SCHEMA, "additionalProperties": False}
+
+QWEN_TOOL_CALL_BEGIN = "<tool_call>\n<function=no_argument_tool>\n"
+QWEN_TOOL_CALL_END = "\n</function>\n</tool_call>"
+
+
+@pytest.mark.parametrize("json_format", XML_FORMATS)
+@pytest.mark.parametrize(
+    "schema",
+    [NO_ARGUMENT_SCHEMA, CLOSED_NO_ARGUMENT_SCHEMA],
+    ids=["implicit_additional_properties", "explicit_additional_properties"],
+)
+def test_no_argument_object_rejects_whitespace(json_format: str, schema: dict):
+    grammar = _json_schema_to_ebnf(schema, json_format=json_format)
+
+    assert _is_grammar_accept_string(grammar, "")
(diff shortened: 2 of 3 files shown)


percymcn:
The diagnosis here is right and the failure mode is a real one — a self-loop with no non-whitespace successor is exactly the shape that produces the vllm#50989 stall, and returning `Empty()` so the enclosing tag's end pattern takes over is the minimal fix.

One additive note, on the exclusion clause rather than the fix. The PR justifies leaving open objects alone with:

> open objects (bare `{"type": "object"}`, `additionalProperties: true` / `{...}`) can genuinely carry properties, so neither is rewritten

That is two claims, and measured on `xgrammar==0.2.5` they come apart. Building the structural tag for a single tool and hashing the emitted grammar (`qwen_3_coder`, `tool_choice="required"`):

| tool `parameters` | grammar md5 | len |
|---|---|---|
| `{"type":"object","properties":{},"additionalProperties":false}` (this PR's target) | `9308ea087981` | 2465 |
| `{"type":"object"}` (excluded) | **`9308ea087981`** | 2465 |
| `{"type":"object","uniqueItems":true}` | **`9308ea087981`** | 2465 |
| `{"type":"object","required":["a"]}` | **`9308ea087981`** | 2465 |
| `{"type":"object","additionalProperties":true}` (excluded) | `632c1ae3e5dd` | 2750 |
| `{"type":"object","properties":{"a":{"type":"string"}},"required":["a"]}` (control) | `85bd9bed1cc6` | 2516 |

`additionalProperties: true` really does compile to something different, so that half of the exclusion is well founded. A **bare `{"type":"object"}` compiles byte-identically to the property-less object this PR is fixing** — so whatever happens to one happens to the other, and after the patch the guard would rewrite one of two identical-grammar inputs.

Scope, stated plainly: I compared emitted grammars, not token-level decoding, so I have not reproduced the stall itself for the bare case — the claim is only that these inputs are indistinguishable to the converter at the point the guard runs.

The underlying reason the clause reads true is that it is a statement about what the schema *means*, while the condition the guard needs is a property of what it *compiles to*. That gap is visible in the plain JSON path too, where the same shapes diverge in the opposite direction:

```python
import json, xgrammar as xgr
def root(s):
    g = str(xgr.Grammar.from_json_schema(json.dumps(s)))
    return [l for l in g.split("\n") if l.startswith("root ::=")][0]

root({"type": "object"})                        # root ::= ((basic_object))          any object
root({"type": "object", "uniqueItems": True})   # root ::= (("{" "}") | ("{" [ \n\t]* "}"))
root({"type": "object", "frobnicate": 1})       # root ::= (("{" "}") | ("{" [ \n\t]* "}"))
root({"type": "object", "title": "t"})          # root ::= ((basic_object))          any object
root({"type": "object", "minProperties": 1})    # RuntimeError
```

A bare object means *any object*, and the moment it carries any keyword the converter does not model it compiles to "the empty object and nothing else". The `frobnicate` row is the control: this is about the **absence of a shape** (`properties` / `additionalProperties` / `$ref` / a combinator), not about any particular keyword — `title`/`description` leave it alone, and adding `additionalProperties: true` restores the correct grammar for every one of them.

`required` is the sharpest instance, because the accepted set ends up disjoint from the schema's — it permits exactly the document the schema forbids and rejects every document it requires:

| | `{}` | `{"a":1}` | `{"a":"x"}` | `{"b":2}` | `{"a":1,"b":2}` | `{"x":"y"}` |
|---|---|---|---|---|---|---|
| JSON Schema | ✗ | ✓ | ✓ | ✗ | ✓ | ✗ |
| `Grammar.from_json_schema` | ✓ | ✗ | ✗ | ✗ | ✗ | ✗ |

That is a separate bug from this PR's and I am not suggesting it belongs here — only that both fall out of the same "can this object hold a property?" question being answered from the schema rather than from the grammar, so a shared predicate would cover both.

One test-design note: every fixture in `test_empty_xml_object.py` uses the property-less spelling (`properties: {}` with `additionalProperties: false`, or `$ref` to one). Since a bare `{"type":"object"}` currently compiles to the same grammar, a regression test built on those fixtures passes whichever way the exclusion goes — the discriminating fixture is a bare `{"type": "object"}` tool parameter, which costs one line.

PR #822, opened 2026-08-08
fix: preserve JSON escapes in length-constrained strings

### LINKED ISSUE #800: JSON Schema maxLength/minLength/pattern on strings drops the escape branch — \n, \", \\, \uXXXX become unemittable
## Summary

When a JSON Schema string field carries **any** of `maxLength`, `minLength`, or `pattern`, the compiled grammar substitutes a length-bounded string production that **drops the JSON escape branch entirely**. The model can no longer emit any RFC 8259 escape sequence — `\"`, `\\`, `\/`, `\b`, `\f`, `\n`, `\r`, `\t`, `\uXXXX` — inside the constrained string.

A plain `{"type": "string"}` accepts all escapes correctly; adding `"maxLength": 8000` (or any length/pattern constraint) makes every escaped payload unreachable.

Verified present in **0.1.33** and still present in **0.2.4** (latest version with cp311/cp312 wheels; 0.2.5 currently publishes cp310-only wheels).

## Impact

Any constrained-decoding deployment that puts a length bound on a prose field silently loses:

- paragraph breaks (`\n` unemittable → wall-of-text output),
- literal double quotes (`\"` unemittable). Combined with the raw `"` remaining legal (it closes the string), a model that wants to write a quotation inside a length-constrained string has **no valid continuation except terminating the string** — we have observed production generations truncating mid-sentence exactly at quotation sites for this reason.

## Reproduction

```python
import json
import xgrammar as xgr

tok = AutoTokenizer.from_pretrained("Qwen/Qwen2.5-0.5B")
info = xgr.TokenizerInfo.from_huggingface(tok, vocab_size=tok.vocab_size)
compiler = xgr.GrammarCompiler(info)

def accepts(schema: dict, payload: str) -> bool:
    cg = compiler.compile_json_schema(json.dumps(schema))
    m = xgr.GrammarMatcher(cg)
    return m.accept_string(json.dumps(payload)) and m.is_completed()

plain = {"type": "string"}
capped = {"type": "string", "maxLength": 8000}

for payload in ["hello", "a\nb", "a\tb", "a\\b", 'a"b']:
    print(f"{payload!r:12}  plain={accepts(plain, payload)}  maxLength={accepts(capped, payload)}")
```

Output on 0.2.4 (identical on 0.1.33):

```
'hello'       plain=True  maxLength=True
'a\nb'        plain=True  maxLength=False
'a\tb'        plain=True  maxLength=False
'a\\b'        plain=True  maxLength=False
'a"b'         plain=True  maxLength=False
```

`minLength` alone, and `minLength`+`maxLength`, reproduce identically.

## Mechanism

`_test_debug_print_grammar` shows the substitution. Plain `{"type": "string"}` uses the standard production with the escape branch:

```
basic_escape ::= (([\"\\/bfnrt]) | ("u" [A-Fa-f0-9] [A-Fa-f0-9] [A-Fa-f0-9] [A-Fa-f0-9])) ...
basic_string_sub ::= (("\"") | ([^\0-\x1f\"\\\r\n] basic_string_sub) | ("\\" basic_escape basic_string_sub)) ...
```

With `maxLength: 8000` the string is expanded into chained per-character rules that only admit the unescaped character class — the `"\\" basic_escape` alternative is absent from every rule:

```
root ::= (("\"" root_130 "\""))
root_2 ::= ("" | ([^\"\\\r\n] root_3))
root_3 ::= ("" | ([^\"\\\r\n] root_4))
...
```

## Expected behavior

Length-constrained string productions should keep the escape branch (counting an escape sequence as one character of the length budget, per JSON Schema semantics where `maxLength` counts code points of the decoded string).

## Environment

- xgrammar 0.1.33 and 0.2.4 (pip wheels), Python 3.11, macOS arm64
- Repro is tokenizer-agnostic (also confirmed with a 150k-vocab Qwen3-family tokenizer)

SyaOtiLan:
## Summary

- Preserve valid JSON escape sequences in strings constrained by `minLength` or `maxLength`.
- Count one complete JSON escape sequence as one character in the repetition budget.
- Reject unescaped U+0000-U+001F control characters in the length-constrained path.

Fixes #800.

The `pattern` path on current main already uses the JSON-string-aware regex expression and retains escapes. This change fixes the remaining dedicated `minLength`/`maxLength` path.

## Testing

- `pytest tests/python/test_json_schema_converter.py -k min_max_length -q` - 12 passed
- `pytest tests/python/test_json_schema_converter.py -q` - 529 passed
- `pytest tests/python/test_json_schema_direct_converter.py -q` - 26 passed
- Ruff, Black, isort, clang-format, line-ending, and whitespace checks passed for the changed files

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -2370,7 +2370,56 @@ int32_t JSONSchemaConverter::GenerateString(const StringSpec& spec, const std::s
   if (spec.min_length != 0 || spec.max_length != -1) {
     int32_t unescaped_character =
         builder_.AddCharacterClass({{0, 0x1f}, {'"', '"'}, {'\\', '\\'}}, true);
-    int32_t escaped_character = Sequence({ByteString("\\"), RuleRef(kBasicEscape)});
+    int32_t hexadecimal_character =
+        builder_.AddCharacterClass({{'A', 'F'}, {'a', 'f'}, {'0', '9'}});
+    int32_t non_surrogate_leading_hex =
+        builder_.AddCharacterClass({{'0', '9'}, {'A', 'C'}, {'E', 'F'}, {'a', 'c'}, {'e', 'f'}});
+    int32_t surrogate_leading_hex = builder_.AddCharacterClass({{'D', 'D'}, {'d', 'd'}});
+
+    int32_t short_escape = Sequence(
+        {ByteString("\\"),
+         builder_.AddCharacterClass(
+             {{'"', '"'},
+              {'\\', '\\'},
+              {'/', '/'},
+              {'b', 'b'},
+              {'f', 'f'},
+              {'n', 'n'},
+              {'r', 'r'},
+              {'t', 't'}}
+         )}
+    );
+    int32_t non_surrogate_unicode_escape = Sequence(
+        {ByteString("\\u"),
+         Choice(
+             {Sequence(
+                  {non_surrogate_leading_hex,
+                   hexadecimal_character,
+                   hexadecimal_character,
+                   hexadecimal_character}
+              ),
+              Sequence(
+                  {surrogate_leading_hex,
+                   builder_.AddCharacterClass({{'0', '7'}}),
+                   hexadecimal_character,
+                   hexadecimal_character}
+              )}
+         )}
+    );
+    int32_t surrogate_pair_escape = Sequence(
+        {ByteString("\\u"),
+         surrogate_leading_hex,
+         builder_.AddCharacterClass({{'8', '9'}, {'A', 'B'}, {'a', 'b'}}),
+         hexadecimal_character,
+         hexadecimal_character,
+         ByteString("\\u"),
+         surrogate_leading_hex,
+         builder_.AddCharacterClass({{'C', 'F'}, {'c', 'f'}}),
+         hexadecimal_character,
+         hexadecimal_character}
+    );
+    int32_t escaped_character =
+        Choice({short_escape, non_surrogate_unicode_escape, surrogate_pair_escape});
     int32_t character = Choice({unescaped_character, escaped_character});
     int32_t body = Repeat(rule_name + "_characters", character, spec.min_length, spec.max_length);
     return Sequence({ByteString("\""), body, ByteString("\"")});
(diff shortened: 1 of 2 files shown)


Seven-Streams:
Thanks for the contribution! Please check whether this comment makes sense:

> 1. Surrogate pairs count as two characters. Both reuse kBasicEscape, whose \uXXXX form is just u[A-Fa-f0-9]{4}. So "\uD83D\uDE00" — one code point, U+1F600 — burns two repetitions, while the same emoji written as raw UTF-8 burns one. The same value passes or fails maxLength: 1 depending on how it's spelled. Issue #800's own "Expected behavior" says maxLength counts code points of the decoded string, so this is precisely the part neither PR delivers. It's specific to the length-bounded path, since that's the only one that counts.
> 2. Lone surrogates are accepted. Any four hex digits are legal, including \uD800–\uDFFF unpaired, so the grammar admits strings that can't be decoded to valid UTF-8. Inherited from kBasicEscape — the unbounded path has it too — so not a regression either PR introduces, just one neither closes.

SyaOtiLan:
Thanks — both observations make sense.

The first is a real gap in #822. The length-bounded rule currently repeats `\` plus `kBasicEscape`, so each `\uXXXX` escape consumes one repetition. A valid surrogate pair such as `\uD83D\uDE00` therefore consumes two repetitions, even though it decodes to the same single code point as the raw UTF-8 character `😀`. This makes `minLength` / `maxLength` depend on how the same decoded value is serialized.

The second observation also accurately describes the current `kBasicEscape` behavior. One nuance is that RFC 8259's ABNF permits unpaired surrogate escapes while warning that the resulting values are not interoperable. Rejecting them would therefore be stricter than the current unbounded-string grammar, and this behavior was not introduced by #822.

The two points are coupled in the length-bounded production: if a valid surrogate pair is added as a one-character alternative while individual surrogate escapes remain valid alternatives, the same input can still be parsed either as one pair or as two lone escapes. In that case, `minLength` can remain incorrect.

For strings with `minLength` or `maxLength`, my proposed scoped fix is to:

- count each JSON short escape (`\"`, `\\`, `\/`, `\b`, `\f`, `\n`, `\r`, and `\t`) and each non-surrogate Unicode escape (`\u0000`–`\uD7FF` or `\uE000`–`\uFFFF`) as one repetition / decoded code point;
- consume a high-surrogate escape (`\uD800`–`\uDBFF`) only when it is immediately followed by a low-surrogate escape (`\uDC00`–`\uDFFF`), with the complete pair counting as one repetition / decoded code point;
- reject unpaired high- or low-surrogate escapes in this length-bounded production;
- leave `kBasicEscape` and the unbounded-string production unchanged.

I can add raw-vs-escaped non-BMP length tests and unpaired-surrogate cases along with that change. Does this scope sound right, or would you prefer surrogate validity to be handled consistently in a separate change?

PR #828, opened 2026-08-11
fix: allow JSON escapes and count code points in length-bounded strings

### LINKED ISSUE #800: JSON Schema maxLength/minLength/pattern on strings drops the escape branch — \n, \", \\, \uXXXX become unemittable

filipsajdak:
A string schema carrying `minLength` or `maxLength` takes a separate lowering from the unbounded one, and that path had two defects. Two commits, reviewable independently.

**`a87833a` -- allow JSON escapes.** The length-bounded character class had no escape alternative:

```
[^"\\\r\n]{min,max}
```

so `\"`, `\\`, `\n` and `\uXXXX` were unrepresentable inside a bounded string. Under constrained decoding a model that needs an escape finds every continuation masked, emits whatever the grammar still allows, and never closes the string -- the generation runs to the token limit. The class also gains the C0 range, which RFC 8259 forbids unescaped and this path had been admitting.

This is the same fix as #822, which is the earlier PR.

**`e14ff8d` -- count code points, not escapes.** `basic_escape` treats every `\uXXXX` alike, so a surrogate pair (one code point spelled as two escapes) cost two repetitions while the raw UTF-8 spelling cost one -- `maxLength: 1` accepted one spelling of U+1F600 and rejected the other. Unpaired `\uD800` was also legal, admitting strings with no valid UTF-8 decoding. The `\u` forms are now enumerated instead of delegating to `basic_escape`: a non-surrogate `\uXXXX` is one repetition, a high+low pair is one repetition, and a lone surrogate matches nothing.

This is the part #800's "Expected behavior" asks for and the escape fix alone does not deliver. The unbounded path is unchanged -- only the bounded path counts characters.

**Tests.** `test_min_max_length_allows_escapes` passes with either fix. The two surrogate tests fail with the escape fix alone (`assert False == True` for the pair, `assert True == False` for the lone surrogate), so they pin the second commit specifically. 547 tests pass across `test_json_schema_converter.py` and `test_json_schema_direct_converter.py`.

Fixes #800

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -2368,14 +2368,46 @@ int32_t JSONSchemaConverter::GenerateString(const StringSpec& spec, const std::s
   }
   // Check for length constraints
   if (spec.min_length != 0 || spec.max_length != -1) {
-    // A length bound limits HOW MANY characters a string has, not which ones.
-    // Without the escape branch \", \\, \n and \uXXXX are unrepresentable, so a
-    // model that needs one cannot close the string and runs to the token limit.
+    // A length bound counts characters, so each repetition must be exactly one
+    // character: an unescaped code point, a short escape, a non-surrogate
+    // \uXXXX, or a surrogate PAIR -- one code point spelled as two escapes.
+    // kBasicEscape cannot be reused here: it treats every \uXXXX alike, which
+    // would make a pair cost two and admit unpaired surrogates.
     int32_t normal_character = builder_.AddCharacterClass(
         {{0, 0x1f}, {'"', '"'}, {'\\', '\\'}, {'\r', '\r'}, {'\n', '\n'}}, true
     );
-    int32_t character =
-        Choice({normal_character, Sequence({ByteString("\\"), RuleRef(kBasicEscape)})});
+    int32_t short_escape = builder_.AddCharacterClass(
+        {{'"', '"'},
+         {'\\', '\\'},
+         {'/', '/'},
+         {'b', 'b'},
+         {'f', 'f'},
+         {'n', 'n'},
+         {'r', 'r'},
+         {'t', 't'}}
+    );
+    int32_t hex = builder_.AddCharacterClass({{'0', '9'}, {'A', 'F'}, {'a', 'f'}});
+    // Surrogates are U+D800-U+DFFF, i.e. "D" followed by 8-F.
+    int32_t hex_but_d =
+        builder_.AddCharacterClass({{'0', '9'}, {'A', 'C'}, {'E', 'F'}, {'a', 'c'}, {'e', 'f'}});
+    int32_t d = builder_.AddCharacterClass({{'D', 'D'}, {'d', 'd'}});
+    int32_t below_surrogates = builder_.AddCharacterClass({{'0', '7'}});
+    int32_t high_surrogate = builder_.AddCharacterClass({{'8', '9'}, {'A', 'B'}, {'a', 'b'}});
+    int32_t low_surrogate = builder_.AddCharacterClass({{'C', 'F'}, {'c', 'f'}});
+    int32_t bmp_escape = Sequence(
+        {ByteString("u"),
+         Choice({Sequence({hex_but_d, hex, hex, hex}), Sequence({d, below_surrogates, hex, hex})})}
+    );
+    int32_t surrogate_pair_escape = Sequence(
+        {ByteString("u"), d, high_surrogate, hex, hex, ByteString("\\u"), d, low_surrogate, hex, hex
+        }
+    );
+    int32_t character = Choice(
+        {normal_character,
+         Sequence({ByteString("\\"), short_escape}),
+         Sequence({ByteString("\\"), bmp_escape}),
+         Sequence({ByteString("\\"), surrogate_pair_escape})}
+    );
     int32_t body = Repeat(rule_name + "_characters", character, spec.min_length, spec.max_length);
     return Sequence({ByteString("\""), body, ByteString("\"")});
   }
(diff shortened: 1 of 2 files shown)


Seven-Streams:
Thanks for the contribution! Please check whether this comment makes sense:

> 1. Surrogate pairs count as two characters. Both reuse kBasicEscape, whose \uXXXX form is just u[A-Fa-f0-9]{4}. So "\uD83D\uDE00" — one code point, U+1F600 — burns two repetitions, while the same emoji written as raw UTF-8 burns one. The same value passes or fails maxLength: 1 depending on how it's spelled. Issue #800's own "Expected behavior" says maxLength counts code points of the decoded string, so this is precisely the part neither PR delivers. It's specific to the length-bounded path, since that's the only one that counts.
> 2. Lone surrogates are accepted. Any four hex digits are legal, including \uD800–\uDFFF unpaired, so the grammar admits strings that can't be decoded to valid UTF-8. Inherited from kBasicEscape — the unbounded path has it too — so not a regression either PR introduces, just one neither closes.

btw it seems to be identical to #822?

filipsajdak:
Thanks for reviewing.

**Both points hold.** `basic_escape` is `escaped_character | unicode_escape` with `unicode_escape ::= "u" hex hex hex hex` (`cpp/json_schema_converter.cc` L1868-L1876), so nothing pairs surrogates anywhere:

1. A pair costs two repetitions while the raw UTF-8 spelling costs one, so `maxLength: 1` accepts one spelling of U+1F600 and rejects the other.
2. `\uD800` unpaired is accepted, so the grammar admits strings that have no valid UTF-8 decoding.

I confirmed this against a build rather than by reading. With the escape fix alone, `maxLength: 1` on `"😀"` fails `assert False == True`, and `"\ud800"` fails `assert True == False`.

**And yes, #822 is the same fix, and it is the earlier PR.** I have split this one so that is visible:

- `a87833a` is the escape fix, equivalent to #822. Its test passes on either branch.
- `e14ff8d` is the part neither PR delivered: the length-bounded path now counts code points, so a surrogate pair is one repetition and a lone surrogate matches nothing. That required not reusing `basic_escape` -- treating every `\uXXXX` alike is exactly what makes a pair cost two.

The unbounded path still accepts lone surrogates. Only the bounded path counts characters, so I left it alone rather than widen the change; happy to do that separately if you want it.

If you would rather take #822 as the base, say so and I will close this -- the surrogate commit can then go on top as its own PR.

Seven-Streams (approved):

PR #834, opened 2026-08-14
fix(json_schema): make prefixItems entries positional per Draft 2020-12

### LINKED ISSUE #824: [Bug] prefixItems without minItems requires all prefix elements
### Checklist

### Describe the bug

`prefixItems` without `minItems` incorrectly requires every declared prefix position to be present. With `strict_mode=False`, arrays shorter than the `prefixItems` list—including the empty array—are rejected, even though Draft 2020-12 defines `prefixItems` as positional validation and does not use it to constrain array length.

This is separate from strict mode's documented treatment of additional items: with `strict_mode=False`, additional items are accepted correctly, but shorter arrays are still rejected.

### Reproduction

```python
import xgrammar as xg

schema = {
    "type": "array",
    "prefixItems": [
        {"type": "string"},
        {"type": "integer"},
    ],
}

chars = 'abcdefghijklmnopqrstuvwxyz0123456789"{}[]:, '
alphabet = [b""] + [c.encode() for c in chars]
ti = xg.TokenizerInfo(
    encoded_vocab=alphabet,
    vocab_type=xg.VocabType.RAW,
    vocab_size=len(alphabet),
    stop_token_ids=[0],
)
tokens = {c: i + 1 for i, c in enumerate(chars)}
compiler = xg.GrammarCompiler(ti)


def accepts(text):
    compiled = compiler.compile_json_schema(
        schema,
        strict_mode=False,
        any_whitespace=True,
    )
    matcher = xg.GrammarMatcher(
        compiled,
        terminate_without_stop_token=True,
    )
    for char in text:
        if not matcher.accept_token(tokens[char]):
            return False
    return matcher.is_terminated()


for text in ('[]', '["a"]', '["a",1]', '["a",1,true]'):
    print(text, accepts(text))
```

Observed output:

```text
[] False
["a"] False
["a",1] True
["a",1,true] True
```

Expected output:

```text
[] True
["a"] True
["a",1] True
["a",1,true] True
```

Reference validation with `jsonschema`:

```python
from jsonschema import Draft202012Validator

for instance in [[], ["a"], ["a", 1], ["a", 1, True]]:
    print(Draft202012Validator(schema).is_valid(instance))
# True, True, True, True
```

The same rejection of empty and shorter arrays also occurs with `strict_mode=True`; strict mode only explains why an additional third item is rejected in that configuration.

Root-cause hypothesis: `ArraySpec.min_items` defaults to zero, but `GenerateArray` unconditionally concatenates every prefix-item rule. This effectively treats the number of `prefixItems` entries as an implicit `minItems` value. The relevant v0.2.5 implementation is in [`GenerateArray`](https://github.com/mlc-ai/xgrammar/blob/v0.2.5/cpp/json_schema_converter.cc#L2268-L2294).

Draft 2020-12 §10.3.1.1 states that `prefixItems` does not constrain array length; positions should only become required through `minItems` or another applicable constraint.

### Environment

- OS: Windows 11 Pro
- Python: 3.12.10
- xgrammar: 0.2.5 (official Windows wheel from the 2026-07-22 v0.2.5 CI build)
- Compiler options: `strict_mode=False`, `any_whitespace=True`, `any_order=False`
- Reference validator: `jsonschema.Draft202012Validator`

feiiiiii5:
## Root Cause
`GenerateArray` built the prefix as `Sequence(prefix_elements)`, forcing **every** `prefixItems` entry to appear. Per JSON Schema Draft 2020-12, `prefixItems` entries are positional — an instance may validly end after any prefix position (subject to `minItems`) — so arrays shorter than the prefix were wrongly rejected (`[]` and `["a"]` for a two-item prefix), and with a naive optional-chain fix the additional-items tail could absorb prefix positions (issue #824).

## Fix
Build the array content as an **alternation of the full prefix (plus the additional tail) and every shorter truncation**, keeping the first `min(minItems, n)` entries mandatory:

```
content = full_prefix (+ additional tail)*
        | truncation ending at each position in [min(minItems, n), n)
```

Additional items are only reachable after the **complete** prefix, so a mismatch inside the prefix range can never be rescued by the additional-items path. Truncations are plain sequences (no optional-chain backtracking), which keeps the grammar unambiguous.

## Test
- 3 new regression tests in `tests/python/test_json_schema_converter.py`:
  - `test_prefix_items_truncated_prefix_accepted`: the issue's accept-set matrix (1111 for `[]`, `["a"]`, `["a",1]`, `["a",1,true]` with `items: {}`), using the discriminating empty/one-element fixtures.
  - `test_prefix_items_min_items_forces_prefix_head`: `minItems` still forces the head of the prefix.
  - `test_prefix_items_no_additional_items_allows_shorter`: `items: false` forbids elements beyond the prefix but must not force the full length.
- Updated 3 existing tests that pinned the old behavior (`test_non_strict`, `test_array_schema` additionalItems/unevaluatedItems cases, `test_array_schema_min_max`): shorter prefixes now accepted, wrong-position values still rejected.
- Local: full `tests/python` suite **3253 passed** (694 skipped, HF_TOKEN-gated), `ruff check` clean, black 24.1.0 / clang-format 19.1.7 clean.

## Diff scope
2 files, +115/-20 (`cpp/json_schema_converter.cc` `GenerateArray` block; `tests/python/test_json_schema_converter.py`).

## AI Disclosure
Root-cause analysis and the first optional-chain draft were AI-assisted; that draft over-accepted (additional items absorbing prefix positions), which was caught by the regression matrix and replaced with the truncation-alternation construction. The final patch was built and verified locally (cmake+ninja, full python suite) and human-reviewed before submission.

Fixes #824

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -2426,13 +2426,9 @@ int32_t JSONSchemaConverter::GenerateArray(const ArraySpec& spec, const std::str
     return spec.min_items == 0 ? Choice({nonempty, empty_array}) : nonempty;
   }
 
-  // Per JSON Schema Draft 2020-12, prefixItems entries are positional: an
-  // instance may validly end after any prefix position (subject to minItems),
-  // but every element inside the prefix range must match its prefix schema —
-  // additional items are only allowed after the full prefix. Build the array
-  // content as an alternation of the full prefix (plus the additional tail)
-  // and every shorter truncation, so shorter arrays are accepted without
-  // letting the additional tail absorb prefix positions (issue #824).
+  // Per Draft 2020-12, prefixItems entries are positional: the instance may
+  // end after any prefix position (subject to minItems), and additional items
+  // are only allowed after the full prefix (issue #824).
   size_t mandatory_count = static_cast<size_t>(std::min<int64_t>(
       std::max<int64_t>(0, spec.min_items), static_cast<int64_t>(item_rule_ids.size())
   ));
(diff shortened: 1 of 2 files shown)


cpp/json_schema_converter.cc
```diff
@@ -2426,31 +2426,70 @@ int32_t JSONSchemaConverter::GenerateArray(const ArraySpec& spec, const std::str
     return spec.min_items == 0 ? Choice({nonempty, empty_array}) : nonempty;
   }
 
+  // Per JSON Schema Draft 2020-12, prefixItems entries are positional: an
+  // instance may validly end after any prefix position (subject to minItems),
+  // but every element inside the prefix range must match its prefix schema —
```
Seven-Streams: simplify the comment

cpp/json_schema_converter.cc
```diff
@@ -2426,31 +2426,70 @@ int32_t JSONSchemaConverter::GenerateArray(const ArraySpec& spec, const std::str
...
+        Sequence({middle_separator, RuleRef(additional_rule_id)}),
+        static_cast<int32_t>(minimum_additional),
+        spec.max_items == -1
+            ? -1
+            : static_cast<int32_t>(spec.max_items - static_cast<int64_t>(item_rule_ids.size()))
+    ));
+  }
+
+  // Truncation alternatives: the prefix may end at exactly k items for
+  // k in [mandatory, n). The full-prefix alternative above covers k == n.
```
Seven-Streams: please flatten the grammar to the form like:

```
rule1 ::= terminal1 rule2
rule2 ::= "" | terminal2 rule3
...
```
instead of several parallel rules with the same prefix, which is harmful to efficiency.

cpp/json_schema_converter.cc
feiiiiii5: Done — the block comment is condensed to three lines in 3f99770.

cpp/json_schema_converter.cc
feiiiiii5: Done in 3f99770. The truncation alternation is replaced with a right-recursive chain ending in the additional-items tail, so each position is encoded once:

```text
root_suffix_2 ::= ("" | ([ \n\t]* "," [ \n\t]* basic_boolean root_additional_items{0, -1}))
root_suffix_1 ::= ("" | ([ \n\t]* "," [ \n\t]* basic_integer root_suffix_2))
root_suffix_0 ::= ("" | (basic_string root_suffix_1))
root ::= (("[" [ \n\t]* root_suffix_0 [ \n\t]* "]"))
```

Position 0, when it is not part of the mandatory head, has no separator prefix since it is the first element. The accepted language is unchanged; the full Python suite passes locally (3253 passed).

Seven-Streams (approved):

PR #835, opened 2026-08-14
fix(json_schema): treat bare 'contains' as minContains=1

### LINKED ISSUE #833: [Bug] type:array without items/prefixItems compiles to an only-empty-array grammar when any keyword is present
Follow-up from the analysis in #824 (thanks to @percymcn for the measurement): a `type: "array"` node with **no shape key** compiles to an only-empty-array grammar as soon as almost any other keyword is present.

## Repro

```python
import json
import xgrammar as xgr

for extra in ({}, {"uniqueItems": True}, {"maxItems": 5}, {"contains": {"type": "integer"}}, {"frobnicate": 1}):
    schema = {"type": "array", **extra}
    g = xgr.Grammar.from_json_schema(json.dumps(schema))
    print(extra or "(no keywords)", _is_grammar_accept_string(g, "[]"), _is_grammar_accept_string(g, "[1]"))
```

- `{}` (no keywords) → accepts `[]` **and** `[1]` (any array)
- any keyword (even a nonexistent one like `frobnicate`) → compiles to `root ::= (("[" [ \n\t]* "]"))`, so `[]` becomes the **only** legal document
- `contains` is the sharpest case: it demands at least one matching element, yet the grammar's accept set `{[]}` is exactly the document the schema forbids; `minItems` there raises `RuntimeError` instead

The control (`frobnicate`) shows the trigger is the *absence of a shape key* (`items`/`prefixItems`) rather than any particular keyword — same code path as #824 from the opposite end (there the prefix is all-mandatory, here the array is all-empty). Fixes to the two should be checked against each other so they don't collapse into one.

(Disclosure: analysis below the repro table is based on percymcn's measurements in #824; I only re-ran and filed it.)

feiiiiii5:
## Root Cause
`ParseArray` only honored `minContains`; a bare `contains` was ignored entirely. A shapeless array (`type: "array"` with no `items`/`prefixItems`) carrying `contains` therefore compiled as if unconstrained — and in strict mode (unevaluatedItems=false) as an only-empty-array grammar that accepts `[]`, which is exactly the document `contains` forbids (issue #833).

## Fix
Per Draft 2020-12, `contains` implies `minContains = 1`. Model it with the same min-count approximation already used for `minContains`:

- `contains: <schema>` → `min_items = max(min_items, 1)`. Non-strict mode now accepts arrays with at least one element; strict mode raises the existing unsatisfiable-schema error ("minItems is greater than the number of prefixItems...") instead of compiling a silently inverted grammar.
- `contains: false` (no element may match) cannot be expressed as a count → explicit unsatisfiable-schema error instead of silent miscompilation.
- `contains` with a non-object/non-boolean value → invalid-schema error.

## Test
3 new regression tests in `tests/python/test_json_schema_converter.py`:
- non-strict: `[]` rejected, `[1]` and `[1, 2]` accepted;
- strict: raises the unsatisfiable-schema error (previously compiled to only-`[]`);
- `contains: false` raises instead of compiling silently wrong.
Local: full `tests/python` suite 3253 passed (694 skipped HF_TOKEN-gated); ruff/black 24.1.0/clang-format 19.1.7 clean.

## Diff scope
2 files, +51/-0 (`cpp/json_schema_converter.cc` ParseArray block; `tests/python/test_json_schema_converter.py`).

## AI Disclosure
Analysis and patch were AI-assisted; the strict_mode semantics (documented as unevaluatedItems=false) and the min-count approximation for `contains` were verified against the repo docs and existing `minContains` handling before submission, and the change was human-reviewed. The general shapeless-array behavior for keywords that do not imply elements (`uniqueItems`/`maxItems`/unknown keys) is the documented strict-mode semantic and is left unchanged.

Fixes #833

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -1232,6 +1232,28 @@ Result<ArraySpec, SchemaError> SchemaParser::ParseArray(const picojson::object& 
     }
     spec.min_items = std::max(spec.min_items, schema.at("minContains").get<int64_t>());
   }
+  if (schema.count("contains")) {
+    // Draft 2020-12: a bare 'contains' implies minContains = 1 (at least one
+    // element must match). Previously the keyword was ignored entirely, so a
+    // shapeless array compiled as if unconstrained and, in strict mode, as an
+    // only-empty-array grammar accepting exactly the document the schema
+    // forbids (issue #833). Model it with the same min-count approximation
+    // already used for 'minContains'; 'contains: false' (no element may match)
+    // cannot be expressed as a count and is rejected explicitly.
+    auto contains_value = schema.at("contains");
+    if (!contains_value.is<picojson::object>() && !contains_value.is<bool>()) {
+      return ResultErr<SchemaError>(
+          SchemaErrorType::kInvalidSchema, "contains must be an object or a boolean"
+      );
+    }
+    if (contains_value.is<bool>() && !contains_value.get<bool>()) {
+      return ResultErr<SchemaError>(
+          SchemaErrorType::kUnsatisfiableSchema,
+          "contains: false requires that no element match, which is not supported"
+      );
+    }
+    spec.min_items = std::max(spec.min_items, static_cast<int64_t>(1));
+  }
   if (schema.count("maxItems")) {
     if (!schema.at("maxItems").is<int64_t>() || schema.at("maxItems").get<int64_t>() < 0) {
       return ResultErr<SchemaError>(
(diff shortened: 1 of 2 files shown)


Seven-Streams:
Thanks for your contribution! But i'm in doubt about the necessity

feiiiiii5:
Thanks for taking a look. The necessity argument has three parts:

1. The default is part of the spec: per Draft 2020-12, a bare `contains` implies `minContains: 1`, so `{"type": "array", "contains": {"type": "integer"}}` must reject `[]`. Today the keyword is dropped entirely, so the schema compiles as unconstrained.

2. In strict mode the current behavior is inverted, not merely lax: with `unevaluatedItems=false` and a shapeless array, dropping `contains` compiles an only-empty grammar, so the accepted set is exactly the document the schema forbids (issue #833).

3. The modeling is the same count-only approximation the parser already applies to an explicit `minContains` (the `spec.min_items` merge just above this hunk), so bare `contains` and `minContains: 1` compile identically, which is the spec's own equivalence.

The known limitation is that the count-only form does not check which element matches: `["a"]` against `contains: {type: integer}` is accepted because the length bound is satisfied. The exact language (at least one element matching the subschema) is expressible as an elements-chain split around one mandatory match, so if you would rather have the precise construction I can rework the patch that way; it is a larger diff than the count merge.

PR #836, opened 2026-08-14
fix(json_schema): honor additionalProperties value schema with propertyNames

### LINKED ISSUE #826: [Bug] propertyNames discards the sibling additionalProperties value schema; top-level alternation in pattern compiles to an unsatisfiable grammar
### Summary

Two silent defects in the JSON Schema → grammar converter (measured on **0.2.4**). Both compile without error and both produce a grammar that does not match the schema, so nothing surfaces them at request time.

I checked #487 first — both cases reported there (`properties`+`patternProperties` coexistence, and `propertyNames` trailing content) are **fixed** in 0.2.4; I re-ran both repros to confirm. These are different.

To keep this reviewable I've used `Grammar.from_json_schema` + `testing._is_grammar_accept_string` rather than a tokenizer, so every line below runs offline in a second.

---

## 1. The presence of `propertyNames` discards the sibling `additionalProperties` value schema

```python
import json, xgrammar as xgr
from xgrammar import testing

def accepts(schema, s):
    g = xgr.Grammar.from_json_schema(json.dumps(schema))
    return testing._is_grammar_accept_string(g, s)

VALUE = {"type": "object", "properties": {"n": {"type": "number"}},
         "required": ["n"], "additionalProperties": False}

with_pn = {"type": "object", "propertyNames": {"type": "string"}, "additionalProperties": VALUE}
without  = {"type": "object", "additionalProperties": VALUE}

assert accepts(without,  '{"a":"plain"}') is False   # correct
assert accepts(with_pn,  '{"a":"plain"}') is False   # FAILS: returns True
```

The value model is simply gone. It is visible in the emitted grammar:

```
# WITH propertyNames
root_prop ::= ([ \n\t]* basic_string [ \n\t]* ":" [ \n\t]* basic_any)

# WITHOUT (control)
root_addl ::= "{" [ \n\t]* (("\"n\"" [ \n\t]* ":" [ \n\t]* basic_number "")) [ \n\t]* "}"
```

`basic_any` replaces the whole value schema. Declared `properties` are **not** affected — I checked; the damage is confined to the map half.

This holds for `propertyNames: {}` and `propertyNames: {"type": "string"}` too, i.e. for a `propertyNames` that asserts nothing at all.

**Why this matters in practice:** zod 4 emits `propertyNames` for *every* `z.record()`:

```js
z.toJSONSchema(z.record(z.string(), z.object({ n: z.number() })))
// {"type":"object","propertyNames":{"type":"string"},
//  "additionalProperties":{"type":"object","properties":{"n":{...}},...}}
```

So on vLLM/SGLang any zod-generated record reaches the decoder with its entire value model unenforced, and the request succeeds.

**Workaround for users today:** delete an unconstrained `propertyNames` (a JSON key is a string by construction, so it asserts nothing), or express a key constraint as `patternProperties` instead — that spelling enforces both key and value correctly.

**Test-design note:** a fixture whose map value is untyped (`additionalProperties: true`, or a bare `{"type":"object"}`) passes either way. The discriminating fixture needs a *typed* value and an instance that violates it.

---

## 2. A top-level alternation in `pattern` compiles to a grammar that accepts nothing

```python
S = {"type": "object",
     "properties": {"a": {"type": "string", "pattern": "cat|dog"}},
     "required": ["a"], "additionalProperties": False}

assert accepts(S, '{"a":"cat"}') is True   # FAILS: returns False
assert accepts(S, '{"a":"dog"}') is True   # FAILS: returns False
```

Not "matches too much" — the field cannot be generated at all. The grammar shows why:

```
root_prop_0 ::= "\"" "c" "a" "t" | "d" "o" "g" "\""
```

The pattern is interpolated between the two quote characters without a wrapping group, so the top-level `|` binds *across* them: the production reads `"cat` OR `dog"`, and neither branch can close a valid JSON string.

The anchored spelling is hit identically, which is the part that surprised me — `^GET|POST$` fails exactly as `GET|POST` does, so writing the pattern carefully does not help.

**Fix shape:** wrap the converted pattern in a non-capturing group before interpolating. `(?:...)` rather than `(...)` leaves any backreference numbering in the user's pattern untouched. Verified: `(?:cat|dog)` accepts `"cat"` and `"dog"` and rejects `"zzz"`.

**Test-design note:** every pattern in the current test suite that I looked at (`[a-z]+`, `\\d{3}`, character classes) passes either way — the discriminating fixture needs an unparenthesised top-level `|`, and the assertion has to be that a **valid** document is *accepted*, since compilation succeeding is precisely what goes wrong here.

---

## Environment

- xgrammar 0.2.4, Python 3.12, macOS
- Both reproduce with `Grammar.from_json_schema`; no tokenizer or model needed.

Happy to split #2 into its own issue if you'd prefer them tracked separately — they're in the same converter but they're independent.

feiiiiii5:
## Root Cause
In `GenerateObject` Case 1b (no named properties, no patternProperties, `propertyNames` present), the value rule was hardcoded to `basic_any`, silently dropping a typed `additionalProperties` schema. Schemas of the shape `{"propertyNames": {...}, "additionalProperties": VALUE}` — emitted by zod 4 for every `z.record()` — reached the decoder with the entire value model unenforced.

## Fix
Use the `additionalProperties` schema as the value rule when one is present; keep values unconstrained (basic_any) only when additional properties are unconstrained or absent, preserving the existing behavior for bare `propertyNames` schemas.

- `{"propertyNames": {"type": "string"}, "additionalProperties": {"type": "object", "properties": {"n": {"type": "number"}}, "required": ["n"], "additionalProperties": false}}`
  - `{"a":"plain"}` now rejected (was accepted)
  - `{"a":{"n":1}}` accepted; `{"a":{"n":"x"}}` rejected

## Test
- New regression test `test_property_names_preserves_additional_properties_value_schema` (typed value fixture with a violating instance, per the issue's test-design note).
- Local: full `tests/python` suite 3251 passed (694 skipped HF_TOKEN-gated); ruff / black 24.1.0 / clang-format 19.1.7 clean.

## Diff scope
2 files, +27/-2 (`cpp/json_schema_converter.cc` Case 1b propertyNames branch; `tests/python/test_json_schema_converter.py`).

## AI Disclosure
Root-cause analysis and patch were AI-assisted; the initial broader patch (treating `additionalProperties: false` as empty-object-only) regressed the existing `test_object_with_pattern_properties_and_property_names` expectation and was narrowed to the typed-schema case, then verified against the full suite and human-reviewed.

Fixes #826 (defect 1). The second defect in the issue (top-level alternation in `pattern`) no longer reproduces on current main — the FSM `Regex` node and the `AddSubGrammar` fallback both handle it correctly.

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -3018,8 +3018,18 @@ int32_t JSONSchemaConverter::GenerateObject(
         }
       } else {
         int32_t key_rule_id = CreateRule(spec.property_names, rule_name + "_name");
-        int32_t value_rule_id = builder_.GetRuleId(GetBasicAnyRuleName());
-        XGRAMMAR_DCHECK(value_rule_id != -1);
+        // propertyNames constrains only the key; a typed
+        // additionalProperties schema must still constrain the value.
+        // Previously the value was hardcoded to basic_any, silently dropping
+        // the additionalProperties value schema (issue #826). When additional
+        // properties are unconstrained (or absent), values stay unconstrained.
+        int32_t value_rule_id;
+        if (additional_property) {
+          value_rule_id = CreateRule(additional_property, rule_name + "_" + additional_suffix);
+        } else {
+          value_rule_id = builder_.GetRuleId(GetBasicAnyRuleName());
+          XGRAMMAR_DCHECK(value_rule_id != -1);
+        }
         property_choices.push_back(Sequence(
             {beginning_separator,
              FormatOtherProperty(
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -3034,7 +3034,7 @@ int32_t JSONSchemaConverter::GenerateObject(
                  value_rule_id,
                  rule_name,
                  /*rule_name_suffix=*/"pn",
-                 /*schema=*/nullptr
+                 additional_property
              )}
         ));
       }
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -3018,11 +3018,8 @@ int32_t JSONSchemaConverter::GenerateObject(
         }
       } else {
         int32_t key_rule_id = CreateRule(spec.property_names, rule_name + "_name");
-        // propertyNames constrains only the key; a typed
-        // additionalProperties schema must still constrain the value.
-        // Previously the value was hardcoded to basic_any, silently dropping
-        // the additionalProperties value schema (issue #826). When additional
-        // properties are unconstrained (or absent), values stay unconstrained.
+        // propertyNames constrains only the key, so a typed additionalProperties
+        // schema still applies to the value (issue #826).
         int32_t value_rule_id;
         if (additional_property) {
           value_rule_id = CreateRule(additional_property, rule_name + "_" + additional_suffix);
--- a/tests/python/test_function_calling_converter.py
+++ b/tests/python/test_function_calling_converter.py
@@ -1690,6 +1690,78 @@ def test_cohere_property_names_reject_invalid_name_or_wrapper(instance: str):
     _check_cohere_grammar(schema, instance, False)
 
 
+@pytest.mark.parametrize("keyword", ["additionalProperties", "unevaluatedProperties"])
+@pytest.mark.parametrize("nested", [False, True], ids=["root", "nested"])
+@pytest.mark.parametrize(
+    "body, accepted",
+    [
+        ("123", True),
+        ("-1", True),
+        ("oops", False),
+        ("1.5", False),
+        ("true", False),
+        ('"oops"', False),
+    ],
+)
+def test_cohere_property_names_preserve_integer_body(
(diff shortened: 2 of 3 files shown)


cpp/json_schema_converter.cc
```diff
@@ -2968,8 +2968,18 @@ int32_t JSONSchemaConverter::GenerateObject(
         }
       } else {
         int32_t key_rule_id = CreateRule(spec.property_names, rule_name + "_name");
-        int32_t value_rule_id = builder_.GetRuleId(GetBasicAnyRuleName());
-        XGRAMMAR_DCHECK(value_rule_id != -1);
+        // propertyNames constrains only the key; a typed
```
Seven-Streams: simplify the doc and we can merge it then

feiiiiii5: Done — fcdae2f condenses the five-line comment to two lines stating what the code does and the issue it fixes. The code and tests are unchanged.

Seven-Streams (approved):

PR #837, opened 2026-08-14
fix(json_schema): emit nothing for empty XML tool-call objects

### LINKED ISSUE #802: [Bug] XML tool calling loops forever on no-argument tools

feiiiiii5:
## Root Cause
For brace-less objects (the XML tool-calling parameter zone) `GenerateObject` built the empty alternative as the whitespace expression `[ \n\t]*` — a self-loop with no non-whitespace continuation. A tool with `parameters = {"type": "object", "properties": {}, "required": []}` therefore left the matcher consuming whitespace forever once the model entered the zone (Qwen favors `\t` there), burning the whole generation budget — vllm-project/vllm#50989.

## Fix
Emit an empty production for the empty alternative when the object is not wrapped in braces. The surrounding tag template already supplies the whitespace, so the tag's `end` takes over immediately. JSON objects (braces) keep the whitespace variant, and the non-empty XML parameter production is unchanged.

- `{"type": "object", "properties": {}, "required": []}` with `qwen_xml` → `root ::= ("")` (was `root ::= [ \n\t]*`)

## Test
- New: `test_qwen_xml_empty_object_emits_nothing` (root rule contains no whitespace loop) and `test_qwen_xml_nonempty_object_keeps_parameter_zone` (guard against over-fixing).
- Updated `test_json_schema_style_kimi_k3_xml_empty_object`: the empty zone no longer accepts `"\n"`/`"\t"` (the zone emits nothing).
- Updated 3 recorded `xml_object` grammar strings in `test_structural_tag_converter.py` (`"" |` replaces the trailing whitespace alternative).
- Local: full `tests/python` suite 3252 passed (694 skipped HF_TOKEN-gated); ruff / black 24.1.0 / clang-format 19.1.7 clean.

## Diff scope
3 files, +35/-5 (`cpp/json_schema_converter.cc` empty-alternative selection; two python test files).

## AI Disclosure
Root cause, fix, and the recorded-grammar updates were AI-assisted; the intermediate over-broad variant was caught by the structural-tag suite and narrowed to the brace-less case, then verified against the full suite and human-reviewed.

Fixes #802

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -3104,7 +3104,11 @@ int32_t JSONSchemaConverter::GenerateObject(
 
   int32_t result = need_braces ? Sequence({ByteString("{"), content, ByteString("}")}) : content;
   if (could_be_empty) {
-    int32_t empty_content = any_whitespace_ ? WhitespaceExpression() : Empty();
+    // Outside braces (XML parameter zones) an empty object must emit
+    // nothing: the surrounding tag template already supplies the whitespace,
+    // and a whitespace-only alternative would be a self-loop that burns the
+    // whole generation budget on tabs (issue #802).
+    int32_t empty_content = any_whitespace_ && need_braces ? WhitespaceExpression() : Empty();
     int32_t empty_result =
         need_braces ? Sequence({ByteString("{"), empty_content, ByteString("}")}) : empty_content;
     return has_content ? Choice({result, empty_result}) : empty_result;
--- a/tests/python/test_json_schema_converter.py
+++ b/tests/python/test_json_schema_converter.py
@@ -3552,6 +3552,27 @@ def test_prefix_items_no_additional_items_allows_shorter():
     check_schema_with_instance(schema, '["a"]', is_accepted=True)
     check_schema_with_instance(schema, '["a", 1]', is_accepted=True)
     check_schema_with_instance(schema, '["a", 1, true]', is_accepted=False)
+def test_qwen_xml_empty_object_emits_nothing():
+    # Regression for issue #802: a qwen_xml tool-call object with no
+    # properties must emit nothing, letting the structural tag's end
+    # take over, instead of a whitespace-only self-loop that never
+    # leaves the parameter zone.
+    ebnf = _json_schema_to_ebnf(
+        {"type": "object", "properties": {}, "required": []}, json_format="qwen_xml"
+    )
+    root_lines = [line for line in ebnf.splitlines() if line.startswith("root ")]
+    assert len(root_lines) == 1
+    assert "[ \\n\\t]*" not in root_lines[0]
+
+
+def test_qwen_xml_nonempty_object_keeps_parameter_zone():
+    # Guard against over-fixing #802: a qwen_xml object with
+    # properties still emits the parameter-zone production.
+    ebnf = _json_schema_to_ebnf(
+        {"type": "object", "properties": {"a": {"type": "integer"}}, "required": []},
+        json_format="qwen_xml",
+    )
+    assert "<parameter=" in ebnf
 
 
 def test_property_names_preserves_additional_properties_value_schema():
--- a/tests/python/test_json_schema_converter.py
+++ b/tests/python/test_json_schema_converter.py
@@ -3552,6 +3552,8 @@ def test_prefix_items_no_additional_items_allows_shorter():
     check_schema_with_instance(schema, '["a"]', is_accepted=True)
     check_schema_with_instance(schema, '["a", 1]', is_accepted=True)
     check_schema_with_instance(schema, '["a", 1, true]', is_accepted=False)
+
+
 def test_qwen_xml_empty_object_emits_nothing():
     # Regression for issue #802: a qwen_xml tool-call object with no
     # properties must emit nothing, letting the structural tag's end
--- a/tests/python/test_structural_tag_converter.py
+++ b/tests/python/test_structural_tag_converter.py
@@ -197,7 +197,7 @@ xml_string ::= TagDispatch(
   excludes=("</parameter>")
 )
 xml_any ::= ((xml_string) | (basic_array) | (basic_object))
-xml_object ::= (([ \n\r\t]* "<parameter=" xml_variable_name ">" [ \n\r\t]* xml_any [ \n\r\t]* "</parameter>" xml_object_properties{0, -1} [ \n\r\t]*) | ([ \n\r\t]*))
+xml_object ::= ("" | ([ \n\r\t]* "<parameter=" xml_variable_name ">" [ \n\r\t]* xml_any [ \n\r\t]* "</parameter>" xml_object_properties{0, -1} [ \n\r\t]*))
 xml_variable_name ::= (([a-zA-Z_] [a-zA-Z0-9_]*))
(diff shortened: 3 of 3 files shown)


Seven-Streams:
I think we can use `max_whitespace_cnt` to solve the problem.

feiiiiii5:
That direction works and is more permissive than the current patch. Tracing the plumbing: `WhitespaceExpression()` already emits `[ \n\t]{0,N}` when `max_whitespace_cnt` is set (json_schema_converter.cc:1614), and `JSONSchemaToGrammar` passes it through to `XMLToolCallingConverter` — but every entry point defaults it to nullopt today, so tool-calling grammars get the unbounded `[ \n\t]*` that produces the #802 loop.

Two ways to adopt it:

1. Default `max_whitespace_cnt` for the XML tool-calling formats in `JSONSchemaToGrammar` (e.g. fall back to a small bound like 16 when nullopt for `kQwenXML`/`kMiniMaxXML`/...). Fixes the loop by default and keeps bounded whitespace legal inside an empty parameter zone, at the cost of a new implicit default the user did not ask for.

2. Keep the default nullopt and only bound the empty-parameter-zone alternative in `GenerateObject` (emit `[ \n\t]{0,N}` there regardless of the global setting). Smaller blast radius, but introduces a local bound that differs from the surrounding whitespace behavior.

The current patch's trade-off for comparison: an empty parameter zone emits `""`, so a model that naturally writes a newline before the closing tag on a zero-argument tool is rejected rather than tolerated. If you prefer option 1 or 2, tell me the bound you want and I will rework the patch; the regression tests would then assert the bounded form instead of `""`.

feiiiiii5:
Following up on this PR and the discussion in #802 — happy to iterate on the fix shape if maintainers prefer a different approach.

feiiiiii5 (commented):
Rebased onto current `main` today and CI is now red — not a flake, and I am not going to paper over it by re-running. It is a semantic conflict with DeepSeek V4.1 support that landed after this PR was opened, so I want a ruling before I touch the shape of the fix again.

**What CI says at head `b7d992b`.** `tests/python/test_builtin_structural_tag.py::test_deepseek_v4_1_empty_arguments[parameters0]`, `[parameters1]`, `[None]` fail — `3 failed, 4041 passed, 57 skipped` on ubuntu-latest (job 106204659064), the same three on the ubuntu and the macOS legs I read (job 106204659227, `3 failed, 4016 passed`); all `build`, `Build wheels`, `Build Docs` and `Pre-check` jobs are green. Those three tests pass on `main` `0b586e9` (unit_test.yaml run 35543443146, success). So the delta is this branch, and only this branch.

**Mechanism.** The hunk in `cpp/json_schema_converter.cc:3112` turns the brace-less empty-object alternative from `WhitespaceExpression()` into `Empty()`. #885 (DeepSeek V4.1, @Ubospica, merged 2026-09-12) added a test — `tests/python/test_builtin_structural_tag.py:2730-2739`, parametrized over `parameters` = `{}`, `{"type": "object", "properties": {}}`, and `None` — asserting the grammar accepts a tool call whose invoke body is two newlines and nothing else. Those two newlines are matched only by the whitespace alternative of the empty-object production, so deleting it makes the string unparseable. They are the only three tests in the whole suite that depend on that alternative: `3 failed, 4041 passed` at head, and the same three pass on `main`.

**Why I am stopping rather than narrowing again.** With only these two alternatives, "the empty zone must force the model to close the tag" (#802) and "the empty zone must tolerate the whitespace a real DeepSeek decode emits" (#885) cannot both hold — a grammar that accepts unbounded whitespace here cannot force termination, and a grammar that accepts none rejects valid output. The options I see, each with who pays for it:

1. **Bound the empty alternative** instead of deleting it, i.e. `[ \n\r\t]{0,n}` at that site. `WhitespaceExpression()` (`cpp/json_schema_converter.cc:2030-2044`) already produces exactly this shape when `max_whitespace_cnt` is set, and returns the unbounded star when it is unset — so this needs a default `n` in the library for the unset case, which is the part @wangxuw argued in #802 should stay with downstream callers. It would still accept the two-newline body that #885's test feeds in.
2. **Keep `Empty()` and change the DeepSeek side** — have the V4.1 template supply its own whitespace around the invoke body, and update #885's three test cases. That is upstream semantics for that feature and not mine to decide alone.
3. **Close this PR** and treat `max_whitespace_cnt`, as @Seven-Streams first suggested on #802, as the answer to the loop, accepting that a brace-less empty object keeps emitting a whitespace alternative.

I would rather not guess which of these the project wants, because options 1 and 2 change behaviour that other people own. Tell me which one and I will implement it here — including the test updates option 2 implies, if that is the call.

Issue #838, opened 2026-08-14
[Bug] pattern/patternProperties use full-string matching instead of Draft 2020-12 unanchored search


feiiiiii5:
## Summary

Draft 2020-12 §6.3.3 (`pattern`) and `patternProperties` require **implicitly unanchored** matching — semantics equivalent to ECMA-262 `RegExp.test()` (a search, not a full match). The converter treats both as **full-string** matches, so valid documents are rejected.

## Repro (current main, offline)

```python
import json, xgrammar as xgr
from xgrammar.testing import _is_grammar_accept_string

S = {"type": "string", "pattern": "cat"}
g = xgr.Grammar.from_json_schema(json.dumps(S))
print(_is_grammar_accept_string(g, '"cat"'))          # True
print(_is_grammar_accept_string(g, '"concatenate"'))  # False — should be True (unanchored)

S2 = {"type": "object", "patternProperties": {"x": {"type": "integer"}}}
g2 = xgr.Grammar.from_json_schema(json.dumps(S2))
print(_is_grammar_accept_string(g2, '{"x": 1}'))   # True
print(_is_grammar_accept_string(g2, '{"x1": 1}'))  # False — should be True (unanchored)
```

The generated grammar shows the FSM `Regex("x", json_string=true)` node matches the whole content (`patternProperties` keys) and `pattern` strings likewise.

## Considerations

This is a broad semantic change, so I did not open a PR unilaterally:

1. Flipping to unanchored search would accept strictly more documents (the Draft-correct direction), but changes the acceptance surface for every downstream consumer (vLLM/SGLang) that currently writes anchored-looking or partial patterns.
2. Implementation-wise the FSM path could wrap the user pattern as a search (`(?:.|
)*(?:PATTERN).*`-style) with the `json_string` character exclusions applied to the wrapper classes; the CFG fallback (`RegexToEBNF`) would need the same treatment.
3. Anchored patterns (`^...$`) behave identically under both semantics, so only partial-match patterns are affected.

Would the maintainers prefer to adopt unanchored (Draft) semantics, or keep the current full-match behavior and document it as a deliberate deviation?

RanaPriyansh:
I reviewed the converter source but have not reproduced the report or implemented a change. Both regex converters discard anchors, so a text wrapper would lose their boundary constraints.

[...]

PR #841, opened 2026-08-14
fix(json_schema): apply propertyNames to the additional branch beside patternProperties

feiiiiii5:
> **Stacked on #839** — please review/merge that first; this branch contains its commit plus the propertyNames change.

## Root Cause
When `patternProperties` and `propertyNames` coexist, the additional-property branch used the unrestricted key pattern, so keys that violate `propertyNames` (and match no pattern) were accepted. `propertyNames` applies to **all** key names per Draft 2020-12; the converter enforced it only when `patternProperties` was absent.

## Fix
Route the additional branch through the `propertyNames` key rule when present, in both the named-properties path (Case 1a) and the pattern-only path (Case 1b):

- `{"patternProperties": {"^x[0-9]+$": {"type": "integer"}}, "propertyNames": {"pattern": "^[a-z0-9]+$"}, "additionalProperties": {"type": "string"}}`
  - `{"x1": 1}` accepted (pattern key, pattern value schema)
  - `{"y": "ok"}` accepted (name passes propertyNames, value passes additionalProperties)
  - `{"UPPER": "bad"}` **rejected** (was accepted — name violates propertyNames)
- Pattern-matching keys keep their pattern key rule (choice semantics; a static grammar cannot intersect the pattern with propertyNames).

## Test
- New `test_property_names_constrains_additional_keys_with_pattern_properties` (anchored patterns, stable under both current and Draft-unanchored regex semantics, cf. #838).
- Local: full `tests/python` suite 3252 passed (694 skipped HF_TOKEN-gated); ruff / black 24.1.0 / clang-format 19.1.7 clean.

## Diff scope
2 files, +35/-3 (`cpp/json_schema_converter.cc` Case 1a + Case 1b merge sites; `tests/python/test_json_schema_converter.py`).

## AI Disclosure
Root-cause analysis and patch were AI-assisted; the static-grammar intersection limit was verified against the existing choice-semantics construction, and the change was human-reviewed after a full-suite run.

Fixes #841 (self-reported; no pre-existing issue).

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -2896,12 +2896,21 @@ int32_t JSONSchemaConverter::GenerateObject(
             RuleRef(key_rule_id), value_rule_id, rule_name, pattern_suffix, pattern_property.schema
         ));
       }
-      // Merge with existing additionalProperties if present
+      // Merge with existing additionalProperties if present. When
+      // propertyNames also constrains the keys, apply it to the additional
+      // branch so non-pattern keys cannot bypass the name constraint
+      // (issue #841). Pattern-matching keys keep their pattern key rule
+      // (choice semantics; a key matching both cannot be intersected by a
+      // static grammar).
       if (effective_additional) {
         int32_t value_rule_id =
             CreateRule(effective_additional, rule_name + "_" + effective_suffix);
+        int32_t additional_key_rule_id = KeyPatternExpression();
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Closing after two weeks without review to keep the queue tidy — the branch stays up, and I'm happy to revive this if it becomes relevant. Thanks!

PR #847, opened 2026-08-17
test: remove inert JSON schema grammar expectations

### LINKED ISSUE #829: check_schema_with_grammar accepts an expected grammar and never compares it
`check_schema_with_grammar` takes an `expected_grammar_ebnf` argument and never compares it. The
whole body is:

```python
json_schema_ebnf = _json_schema_to_ebnf(schema, any_whitespace=..., indent=..., ...)
# Direct AST construction may reuse rules and print repetition nodes differently from the
# former handwritten EBNF converter. Preserve stable rule-level checks here; language behavior
# is covered by the acceptance/rejection tests in this file and test_json_schema_direct_converter.
assert expected_grammar_ebnf
for rule_name in ("basic_escape", "basic_string", "basic_array", "basic_object", "root"):
    assert f"{rule_name} ::=" in json_schema_ebnf
```

`assert expected_grammar_ebnf` is a truthiness check on the caller's string. The remaining
assertions check that five rule names appear somewhere in the output — true for essentially any
JSON grammar the converter can emit.

The comment shows this was a deliberate loosening during the AST refactor, and the reasoning is
sound: rule reuse and repetition-node printing made exact string comparison brittle. The problem is
what was left behind. `tests/python/test_json_schema_converter.py` still carries several hundred
carefully written expected-EBNF blocks that are now inert, and they read exactly like grammar
assertions to anyone scanning the file.

**Concretely:** in #828 I changed the emitted grammar for every length-bounded string, from

```
root ::= "\"" [^"\\\r\n]{1,10} "\""
```

to

```
root_characters ::= [^\0-\x1f"\\\r\n] | "\\" basic_escape
root ::= "\"" root_characters{1,10} "\""
```

`test_min_max_length` asserts the first form verbatim and passed unmodified against the second. The
full file — 518 tests — passed with the grammar rewritten underneath it. I only noticed because the
result contradicted what I had just printed from the converter; had the two agreed I would have
believed the suite.

I am not proposing exact string comparison be restored. Some options that keep the refactor's
benefit:

1. **Normalise then compare** — parse both sides and compare structurally, so rule naming and
   repetition printing do not matter but the accepted language does.
2. **Compare a canonical serialisation** of the grammar rather than the EBNF text.
3. **Drop the parameter** and delete the dead expectations, making the acceptance tests the honest
   and only contract. Least work, and truthful.
4. **Keep it but rename** to something like `check_schema_grammar_builds`, so the call sites stop
   implying a comparison that does not happen.

Whichever way, the current shape has the failure mode that a check which cannot fail looks
identical to one that passes — and here it sits on top of expectations that still look
authoritative.

Happy to send a PR for whichever direction you prefer.

yu-xin-c:
## Summary

- remove expected EBNF fixtures that stopped being compared after the direct AST migration
- rename the helper to describe its grammar-build smoke-test contract
- parse the generated EBNF through `Grammar.from_ebnf` so the serialization path is meaningfully checked
- preserve all schema and acceptance/rejection coverage

## Rationale

The direct AST converter intentionally reuses and prints grammar nodes differently from the former handwritten EBNF converter. Canonicalizing those representations would still make these tests depend on implementation details, while general CFG language equivalence is not a practical unit-test oracle. Existing acceptance/rejection cases remain the semantic contract.

## Validation

- `black --check tests/python/test_json_schema_converter.py`
- `isort --check-only tests/python/test_json_schema_converter.py`
- `ruff check tests/python/test_json_schema_converter.py`
- `pytest -q tests/python/test_json_schema_converter.py` (518 passed)

Closes #829

### DIFF
--- a/tests/python/test_json_schema_converter.py
+++ b/tests/python/test_json_schema_converter.py
@@ -16,34 +16,9 @@ from xgrammar.testing import (
     _json_schema_to_ebnf,
 )
 
-basic_json_rules_ebnf = r"""basic_escape ::= ["\\/bfnrt] | "u" [A-Fa-f0-9] [A-Fa-f0-9] [A-Fa-f0-9] [A-Fa-f0-9]
-basic_string_sub ::= ("\"" | [^\0-\x1f\"\\\r\n] basic_string_sub | "\\" basic_escape basic_string_sub) (= [ \n\t]* [,}\]:])
-basic_any ::= basic_number | basic_string | basic_boolean | basic_null | basic_array | basic_object
-basic_integer ::= ("0" | "-"? [1-9] [0-9]*)
-basic_number ::= "-"? ("0" | [1-9] [0-9]*) ("." [0-9]+)? ([eE] [+-]? [0-9]+)?
-basic_string ::= ["] basic_string_sub
-basic_boolean ::= "true" | "false"
-basic_null ::= "null"
-basic_array ::= (("[" [ \n\t]* basic_any ([ \n\t]* "," [ \n\t]* basic_any)* [ \n\t]* "]") | ("[" [ \n\t]* "]"))
-basic_object ::= ("{" [ \n\t]* basic_string [ \n\t]* ":" [ \n\t]* basic_any ([ \n\t]* "," [ \n\t]* basic_string [ \n\t]* ":" [ \n\t]* basic_any)* [ \n\t]* "}") | "{" [ \n\t]* "}"
-"""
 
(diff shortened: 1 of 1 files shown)


Seven-Streams:
Thanks for your contribution! I'm wondering whether we can add the exact tests back again?

yu-xin-c:
Thanks. I can add exact coverage back. The current direct-AST serialization differs systematically from the old fixtures in rule reuse, generated rule numbering, and repetition rendering, so I want to confirm the intended scope: should I restore and update every inline expected EBNF fixture to the current serialization, or keep the cleanup and add a smaller set of focused exact-serialization regression tests for representative schemas? I have verified that simply restoring equality makes the first existing fixture fail for these representation changes.

Seven-Streams (approved):

PR #849, opened 2026-08-17
fix(fsm): complement negated character classes over the full codepoint domain

### LINKED ISSUE #848: Negated character class over non-ASCII ranges silently fails to enforce (clamped to 127)
## Summary

Negated character classes (`[^...]`) that exclude non-ASCII characters have compilation issues. xgrammar emits `Negative Character class contains byte greater than 127, clamping to 127.` and then produces a token mask that **allows tokens the regex excludes**. Hence for `[^<non-ASCII ranges>]` the exclusion doesn't actually take effect.

We hit this trying to constrain an LLM's output to a single script (e.g. "never output Bengali/Devanagari/CJK, only Kannada + ASCII") for a multilingual scenario. The generated text still contained the "excluded" scripts, and in some cases the model fell into a degenerate single-token repetition loop.

## Reproduction

```python
import xgrammar as xgr

# Tiny synthetic vocabulary: plain ASCII + one Bengali word.
vocab = ["a", "b", "c", " ", "hello", "world", "0", "1", "কা", "<eos>"]
BENGALI_TOKEN_ID = vocab.index("কা")

tok_info = xgr.TokenizerInfo(vocab, vocab_size=len(vocab), stop_token_ids=[len(vocab) - 1])
compiler = xgr.GrammarCompiler(tok_info)

# Exclude the Bengali Unicode block. Anything else (ASCII, digits, etc.) should stay legal.
bengali_range = "ঀ-ঃঅ-ঌএ-ঐও-নপ-রলশ-হ়-ৄে-ৈো-ৎৗড়-ঢ়য়-ৣ০-৾"
pattern = f"^[^{bengali_range}]*$"

g = xgr.Grammar.from_regex(pattern)
c = compiler.compile_grammar(g)
```

This alone prints, 14 times (once per range segment in the class):

```
Warning: Negative Character class contains byte greater than 127, clamping to 127.
```

Now checking what the compiled grammar actually allows as the very first token —

```python
matcher = xgr.GrammarMatcher(c)
bitmask = xgr.allocate_token_bitmask(1, tok_info.vocab_size)
matcher.fill_next_token_bitmask(bitmask)

def allowed(token_id):
    word = bitmask[0][token_id // 32].item()
    return bool((word >> (token_id % 32)) & 1)

print("Bengali token allowed?", allowed(BENGALI_TOKEN_ID))  # -> True (should be False)
print("'a' allowed?", allowed(vocab.index("a")))             # -> True (correct)
```

**Expected**: `Bengali token allowed?` should print `False` — the regex says that token can never appear.
**Actual**: it prints `True`. The exclusion has no effect.

## Specific to negated classes containing non-ASCII

The same non-ASCII ranges in a **positive** (allowlist) class never hit this at all — 0 warnings, and it behaves correctly, even when the positive class is much larger (we tested unioning 11 different scripts' worth of ranges into one allowlist: 0 warnings, correct enforcement). It's specifically `[^...]` + byte value > 127 that triggers the clamp-to-127 path.

## Environment

- xgrammar: 0.2.3 (installed standalone via pip; also reproduced with 0.2.0, as bundled with vLLM 0.20.1)
- Python: 3.12.3
- torch: 2.13.0+cu130
- transformers: 5.15.0
- OS: Linux (x86_64)

## Why this matters

`[^...]` over non-ASCII is a  way to express "never output script X" for multilingual generation constraints. Right now it silently fails once byte values exceed 127.

OmarFaig:
BuildNegativeCharacterClass computed the complement of a negated character class ([^...]) over a 128-entry ASCII bitset, then unconditionally added an edge accepting every codepoint from kMin2BytesUnicode to kMax4BytesUnicode regardless of what the class excluded. So [^<script>] correctly rejected the intended ASCII exclusions but admitted every non-ASCII codepoint outright, including ones the pattern explicitly excluded — the resulting token mask permitted tokens the regex said could never appear. The "byte greater than 127, clamping to 127" warning was a downstream symptom: the class endpoints are Unicode codepoints, but this function truncated them to uint8_t before the range check, mangling any non-ASCII endpoint before the clamp even fired.

The fix computes the complement over the full codepoint domain [0, 0x10FFFF] instead: collect the class's ranges, sort and merge them, then walk the gaps and emit an FSM edge for each, reusing the existing AddCharacterRange/CodepointToPackedUTF8 helpers the positive-class path already uses for UTF-8 packing. This mirrors the sort-merge-complement shape RepetitionNormalizerImpl::AppendPositiveRanges already uses elsewhere in this file. The clamp and its warning are gone.

New cases in test_grammar_matcher_regex.py cover range rejection, non-ASCII acceptance outside the excluded range, a positive-class regression guard, the ASCII/non-ASCII boundary, the top of the codepoint domain, and a token-mask-level test reproducing the original symptom directly. Two golden FSM/serialization snapshots embedding the old, over-permissive UTF-8 lead-byte range are updated to match.

Fixes #848.


Ubospica:
cc @Seven-Streams Looks good to support this

PR #860, opened 2026-08-23
fix(json_schema): support draft-07 positional items and additionalItems

### LINKED ISSUE #840: [Bug] draft-07 additionalItems is silently ignored (true and schema forms reject valid documents)
## Summary

The draft-07 keyword `additionalItems` is not parsed at all. Under strict mode (the default) its absence happens to mimic `additionalItems: false`, which masks the gap — but `additionalItems: true` and `additionalItems: <schema>` are silently ignored and valid documents are rejected.

## Repro (current main, offline)

```python
import json, xgrammar as xgr

S = {"type": "array", "prefixItems": [{"type": "integer"}], "additionalItems": True}
g = xgr.Grammar.from_json_schema(json.dumps(S))
print(_is_grammar_accept_string(g, "[1, 2]"))   # False — draft-07: valid

S2 = {"type": "array", "prefixItems": [{"type": "integer"}],
      "additionalItems": {"type": "string"}}
g2 = xgr.Grammar.from_json_schema(json.dumps(S2))
print(_is_grammar_accept_string(g2, '[1, "x"]'))  # False — draft-07: valid
```

The existing `test_array_schema` case that uses `"additionalItems": False` passes only because strict mode already closes the array — the keyword itself has no effect.

## Question

XGrammar targets Draft 2020-12 (where `additionalItems` was folded into `items`). Options:

1. Support `additionalItems` as an alias of `items` for draft-07 inputs (accepts more documents; small ParseArray change).
2. Reject the keyword with a clear "unsupported in Draft 2020-12" error instead of silently ignoring it (honest failure for schema authors porting draft-07 schemas).
3. Keep ignoring it (documented deviation).

Happy to implement either 1 or 2 once the direction is confirmed.

feiiiiii5:
Fixes #840

## Description

Draft-07 schemas may write `items` as an **array of schemas** applied positionally (equivalent to 2020-12 `prefixItems`), with `additionalItems` constraining the elements beyond that list. XGrammar previously:

- rejected positional `items` arrays with `"items must be a boolean or an object"`, and
- silently ignored `additionalItems` entirely — so `{"prefixItems": [{"type": "integer"}], "additionalItems": true}` rejected `[1, 2]` though draft-07 calls it valid, and `additionalItems: <schema>` had no effect.

Both are fixed here:

1. `ParseArray` now parses a positional `items` array into `prefix_items`, reusing the existing prefixItems semantics and error messages.
2. `additionalItems` (boolean or schema) is applied after any positional list, matching draft-07: `false` closes the array, `true` leaves extra elements unrestricted, a schema constrains them. When `items` is a single schema/boolean (non-positional), `additionalItems` has no meaning per the draft and is left untouched.
3. Without an explicit `additionalItems`, the strict/non-strict default is preserved so positional `items` behaves exactly like `prefixItems` today.

## Tests

- Extended `schema__err_message__test_array_schema_error_cases` (positional items element type, `additionalItems` type check).
- Added `test_draft07_positional_items_and_additional_items` covering: positional items + `additionalItems` false/true/schema, `prefixItems` + `additionalItems` (the two repro cases from #840), non-strict defaults, and boolean `items`.
- `tests/python/test_json_schema_converter.py`: **521 passed**; adjacent `test_json_schema_direct_converter.py` / `test_grammar_matcher_json_schema.py`: 29 passed (110 skipped = HF-token model tests, unrelated).
- `clang-format` (v19.1.7, project pin) clean.

## Checklist

- [x] Tests added/updated and passing
- [x] Commit formatted with project clang-format config

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -1173,15 +1173,68 @@ Result<ArraySpec, SchemaError> SchemaParser::ParseArray(const picojson::object& 
     }
   }
 
+  // Draft-07 form: "items" may be an array of schemas applied positionally,
+  // equivalent to prefixItems (with additionalItems governing the elements
+  // beyond that list). Parse that positional form into prefix_items so the
+  // rest of this function treats it uniformly with prefixItems.
+  if (schema.count("items") && schema.at("items").is<picojson::array>()) {
+    for (const auto& item : schema.at("items").get<picojson::array>()) {
+      if (item.is<bool>() && !item.get<bool>()) {
+        return ResultErr<SchemaError>(
+            SchemaErrorType::kUnsatisfiableSchema, "items contains false"
+        );
+      } else if (!item.is<picojson::object>()) {
+        return ResultErr<SchemaError>(
+            SchemaErrorType::kInvalidSchema, "items array must only contain objects or booleans"
+        );
+      }
+      auto item_result = Parse(item, "item");
+      if (item_result.IsErr()) return ResultErr(std::move(item_result).UnwrapErr());
+      spec.prefix_items.push_back(std::move(item_result).Unwrap());
+    }
+  }
+
   if (schema.count("items")) {
     auto items_value = schema.at("items");
-    if (!items_value.is<bool>() && !items_value.is<picojson::object>()) {
+    if (!items_value.is<bool>() && !items_value.is<picojson::object>() &&
+        !items_value.is<picojson::array>()) {
       return ResultErr<SchemaError>(
-          SchemaErrorType::kInvalidSchema, "items must be a boolean or an object"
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Friendly follow-up: this PR (draft-07 positional `items` + `additionalItems` support, Fixes #840) is ready for review — 521/521 json-schema converter tests pass, clang-format clean, and it adapts the existing prefixItems machinery (no new architecture). If the CI gate could be approved, happy to address any feedback. No rush otherwise.

feiiiiii5:
Gentle nudge: this one is still ready on my side (rebased, CI green, 2-file diff for draft-07 positional items + additionalItems, fixes #840). Happy to adjust scope or split anything that helps review.

PR #869, opened 2026-08-30
Fix segfaults from invalid arguments, corrupted serialized data and worker-thread errors

Ubospica:
## Summary

Hardens the C++ boundary so that invalid arguments, corrupted or incomplete serialized data, and errors raised on worker threads surface as Python exceptions instead of crashing the process. Every fix comes with a subprocess-verified reproducer turned into a regression test.

### Out-of-range arguments to the public API

- `apply_token_bitmask_inplace(..., indices=[...])`: validate every index against both the logits and the bitmask batch size (`ApplyTokenBitmaskInplaceCPU` only checked the no-indices path, so an out-of-range index wrote through a wild pointer).
- `TokenizerInfo(vocab, vocab_size=N)` with `N < len(vocab)`: reject in the constructor (`token_id_to_sorted_vocab_index_` was allocated with `vocab_size` and written with real token ids).
- `accept_token(padding_id)`: padding ids pass the range check because it uses the padded vocab size, but the special-token warning indexed `decoded_vocab` with them; only decode ids backed by a real token.
- `traverse_draft_tree`: validate that every `retrieve_next_token` / `retrieve_next_sibling` value is in `[-1, num_nodes)` and that the edges form a tree (root has no incoming edge, every other node at most one). Out-of-range values recursed into arbitrary memory; cycles or shared nodes recursed without bound.
- EBNF macro arguments: nested tuples in `ParseMacroValue` bypassed `nest_layer_guard_`, so `Token((((...` overflowed the stack.
- Token ids written in `Token(...)`, `ExcludeToken(...)` and token tag dispatch must exist in the vocabulary. They are used as indices into per-token arrays on compiler worker threads, so they are checked on the calling thread before compilation starts (and skipped defensively on the worker path).
- `GrammarMatcher(override_stop_tokens=[...])`: every id must be in `[0, vocab_size)`; the ids are written into the token bitmask.

### Integer truncation and unbounded recursion in the grammar front-ends

- JSON schema `minLength` / `maxLength` / `minItems` / `minContains` / `maxItems` beyond int32 were truncated when converted to repetition bounds; a negative upper bound made `LegacyHandleRepetitionRange` call `back()` on an empty vector. Maxima beyond int32 are now unbounded, minima beyond it are rejected, and `ExpandRepetitionRange` checks its range.
- The Lark parser had no nesting limit for groups; it now stops at 200 levels (each level uses a few KB of stack; the limit leaves a wide margin below the 1 MB main-thread stack on Windows).
- Structural tag formats are limited to 100 nesting levels; the global recursion limit is far deeper than the native stack allows for the passes over the format tree.
- `picojson::parse` recurses once per nesting level with no limit. All call sites go through `ParseJSON` (`cpp/support/json_parse.h`), which parses with a picojson context that counts every nested array or object in a `RecursionGuard`, so the maximum recursion depth applies while parsing and a too-deep input is reported as a JSON parse error.
- `GrammarBuilder::AddGrammarExpr` checks that the grammar expr data stays addressable by int32 offsets (reachable with a huge `indent`).
- The `xgrammar.testing` FSM helpers check `per_rule_fsms` before indexing it.

### Deserialization of corrupted or incomplete data

The reflection-based deserializer restored CSR arrays, ids and bitsets verbatim and never re-ran the invariants that the constructors guarantee, so corrupted JSON crashed on first use (`root_rule_id`, `grammar_expr_indptr`, FSM edges, mask indices, ...). It also never rebuilt `token_id_to_sorted_vocab_index_`, so compiling any grammar with token edges against a *legitimately* deserialized `TokenizerInfo` segfaulted.

- `AutoDeserializeJSONValue` now calls `std::optional<std::string> Validate() const` on reflection-based types that define it, after all members are deserialized, and reports the message as a `DeserializeFormatError`. `Validate()` is implemented for `Compact2DArray`, `CompactFSM::Impl`, the `CompactFSMWithStartEnd` serialization helper, `Grammar::Impl` and `TokenizerInfo::Impl`.
- `DynamicBitset` deserialization checks the array length before reading the words.
- `CompiledGrammar` deserialization propagates the errors of its `grammar` and `adaptive_token_mask_cache` sub-objects (they were ignored) and checks that every mask matches the tokenizer.
- `TokenizerInfo::DeserializeJSON` rebuilds `token_id_to_sorted_vocab_index_`.

The serialization format is unchanged.

### Errors on worker threads

`ThreadPool::Execute` tasks that threw (a terminated matcher or an out-of-range index in `BatchGrammarMatcher.batch_fill_next_token_bitmask`, or any check failing during multi-threaded compilation) escaped the worker thread and called `std::terminate`. The pool now records the first exception and rethrows it from `Wait()` / `Join()` on the calling thread.

### DIFF
--- a/cpp/grammar_compiler.cc
+++ b/cpp/grammar_compiler.cc
@@ -734,6 +734,12 @@ const std::vector<int32_t>& GrammarMatcherForTokenMaskCache::GetTokenEdgeAccepte
   const auto& sorted_decoded_vocab = tokenizer_info_.GetSortedDecodedVocab();
   int32_t sorted_size = static_cast<int32_t>(sorted_decoded_vocab.size());
   const auto& tid_to_sorted = tokenizer_info_.ImplPtr()->GetTokenIdToSortedVocabIndex();
+  // Out-of-vocabulary ids are rejected before compilation starts (CheckTokenIdsInVocab); guard
+  // here as well because this runs on worker threads, where an out-of-bounds read cannot be
+  // reported as an error.
+  auto sorted_index = [&](int32_t tid) {
+    return tid >= 0 && tid < static_cast<int32_t>(tid_to_sorted.size()) ? tid_to_sorted[tid] : -1;
+  };
 
   bool has_exclude_token = false;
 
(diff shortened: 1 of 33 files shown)


cpp/grammar_compiler.cc
Seven-Streams, line 755: maybe no need to change?

cpp/grammar_compiler.cc
```diff
@@ -734,27 +734,29 @@ const std::vector<int32_t>& GrammarMatcherForTokenMaskCache::GetTokenEdgeAccepte
...
   for (const auto& edge : edges) {
     if (edge.IsToken()) {
       auto info = fsm.GetFsm().GetFsm().GetTokenEdgeInfo(edge.GetAuxIndex());
       for (int32_t i = 0; i < info.Count(); ++i) {
-        int32_t tid = info.TokenIds()[i];
-        XGRAMMAR_DCHECK(tid >= 0 && tid < static_cast<int32_t>(tid_to_sorted.size()));
-        if (tid_to_sorted[tid] >= 0) {
-          tmp_token_edge_accepted_.push_back(tid_to_sorted[tid]);
+        if (int32_t index = sorted_index(info.TokenIds()[i]); index >= 0) {
+          tmp_token_edge_accepted_.push_back(index);
```
Seven-Streams, line 751: maybe no need to change

PR #872, opened 2026-09-01
feat: add canonical recursive Any support for Cohere XML

jasonozuzu-cohere:
Adds canonical recursive Cohere XML support for unrestricted JSON Schema values, correlating raw, json, dict, and list wrappers with their matching body grammars. Also aligns root true and {} behavior and adds coverage for nested values, references, composites, and unrestricted tool parameters

### DIFF
--- a/tests/python/test_builtin_structural_tag.py
+++ b/tests/python/test_builtin_structural_tag.py
@@ -1084,6 +1084,106 @@ def test_cohere_required_accepts_nested_params():
     check_stag_with_instance(structural_tag, named_list_item_output, False)
 
 
+@pytest.mark.parametrize(
+    "function",
+    (
+        pytest.param({"name": "anything"}, id="omitted-parameters"),
+        pytest.param({"name": "anything", "parameters": None}, id="none-parameters"),
+        pytest.param(
+            {
+                "name": "anything",
+                "strict": False,
+                "parameters": {
+                    "type": "object",
+                    "properties": {"only": {"type": "integer"}},
+                    "required": ["only"],
+                    "additionalProperties": False,
+                },
+            },
+            id="strict-false",
+        ),
+    ),
+)
+@pytest.mark.parametrize(
+    "body, accepted",
+    (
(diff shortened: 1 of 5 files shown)


tests/python/test_builtin_structural_tag.py
Seven-Streams, line 1: we need more tests about `strict=False` for the check of `additionalProperties: true`.

Seven-Streams (commented):
Thanks for your contribution! btw, could you double check if this makes:

> Non-identifier keys in nested dict values are rejected
> 
> Under Any, a nested object is now only expressible as a type="dict" wrapper, and its child names go through xml_variable_name ([A-Za-z_][A-Za-z0-9_]*). Keys like my-key, a.b, 0, or non-ASCII keys are rejected.
> 
> Repro via the alignment test's own pipeline: a tool with parameters: {"type": "object"} and arguments {"meta": {"my-key": 1}} renders through cmd5 as
> 
> <cofl:value name="meta" type="dict"><cofl:value name="my-key" type="json">1</cofl:value></cofl:value>
> 
> and the grammar rejects it. melody's cmd5 output parser round-trips this text back to {"meta": {"my-key": 1}}, so such keys are a supported part of model output.
> 
> Before this PR the value was still accepted as type="json">{"my-key":1}, which melody also parses. Now it has no accepted spelling at all. Declared properties are unaffected since they are emitted as literals; only dynamic keys (additionalProperties, Any) hit this.
> 
> Suggested fix: relax the Cohere key rule (name is a quoted attribute, so anything except " would do), or keep the type="json" fallback for Any bodies. Please also add a test with a non-identifier nested key, ideally as a melody alignment case with nested arguments.

tests/python/test_builtin_structural_tag.py
jasonozuzu-cohere: Extended the number of pytest params on the test

jasonozuzu-cohere:
@Seven-Streams good catch! Have refactored to make xml key names for the new format more permissive

Seven-Streams (approved):

PR #880, opened 2026-09-08
fix: preserve JSON escapes in length-constrained strings

feiiiiii5:
## Root cause

`JSONSchemaConverter::GenerateString` lowers a string with `minLength`/`maxLength` to a bounded repetition over a bare negated class:

```cpp
builder_.AddCharacterClass({{'"', '"'}, {'\\\\', '\\\\'}, {'\\r', '\\r'}, {'\\n', '\\n'}}, true)
```

That production has no `"\\\\" basic_escape` alternative, so every RFC 8259 escape (`\\"`, `\\\\`, `\\/`, `\\b`, `\\f`, `\\n`, `\\r`, `\\t`, `\\uXXXX`) is unreachable inside a length-bounded string, while plain `{"type": "string"}` accepts them via `basic_string_sub`. The class also admits unescaped U+0000–U+001F (except CR/LF), which RFC 8259 forbids. Reported in #800; the `pattern` path already uses the JSON-string-aware regex and is unaffected, and `propertyNames` bounds flow through this same function so they are fixed too.

## Fix

Each `Repeat` element now matches exactly one JSON character (JSON Schema maxLength counts characters of the decoded string):

- unescaped code point `[^\\0-\\x1f"\\\\]` (mirrors `AddHelperRules`, closing the C0 hole),
- short escape `\\\\["\\\\/bfnrt]` — one repetition,
- BMP `\\uXXXX` (non-surrogate lead, incl. `D0`–`D7`) — one repetition,
- surrogate pair `\\uD83D\\uDE00` (high `D8`–`DB` + low `DC`–`DF`) — one repetition; lone surrogates match nothing.

`kBasicEscape` is deliberately not reused: it accepts every `\\uXXXX` uniformly, which would make a pair cost two repetitions and admit lone surrogates.

## Tests

- New in `tests/python/test_json_schema_converter.py`:
  - `test_min_max_length_allows_json_escapes` — all short escapes + `\\uXXXX` accepted under `maxLength`/`minLength`/both (issue repro: `a\\nb`, `a\\tb`, `a\\\\b`, `a\\"b`),
  - `test_min_max_length_counts_escape_as_one_character` — `"a\\nb"` needs `minLength: 3`, exceeds `maxLength: 2`; 10 vs 11 escapes against `maxLength: 10`,
  - `test_min_max_length_rejects_unescaped_controls` — raw TAB/LF rejected,
  - `test_min_max_length_counts_surrogate_pair_as_one_character` — raw U+1F600 and `\\uD83D\\uDE00` fit `maxLength: 1`, two pairs do not,
  - `test_min_max_length_rejects_lone_surrogate_escapes` — `\\uD800`/`\\uDC00` rejected, `\\uD7FF`/`\\uE000` accepted.
- Results: `test_json_schema_converter.py` — 527 passed; `test_json_schema_direct_converter.py` — 26 passed. `clang-format`, `ruff`, `isort` clean for the touched lines.

Refs #800

### DIFF
--- a/cpp/json_schema_converter.cc
+++ b/cpp/json_schema_converter.cc
@@ -2370,8 +2370,63 @@ int32_t JSONSchemaConverter::GenerateString(const StringSpec& spec, const std::s
   }
   // Check for length constraints
   if (spec.min_length != 0 || spec.max_length != -1) {
-    int32_t character =
-        builder_.AddCharacterClass({{'"', '"'}, {'\\', '\\'}, {'\r', '\r'}, {'\n', '\n'}}, true);
+    // Each repetition must match exactly one JSON character so the Repeat budget counts
+    // characters (JSON Schema maxLength counts characters of the decoded string). An
+    // unescaped code point is one repetition; one complete escape sequence is also one.
+    // A bare [^"\\\r\n] class here would drop the escape branch entirely (issue #800),
+    // making \n/\"/\\/\\uXXXX unemittable inside length-bounded strings.
+    // kBasicEscape is not reused: it accepts every \uXXXX uniformly, so a surrogate pair
+    // (one code point spelled as two escapes) would cost two repetitions and lone
+    // surrogates would be admitted. Enumerate instead: a BMP escape is one repetition,
+    // a high+low surrogate pair is one repetition, and a lone surrogate matches nothing.
+    // The unescaped class mirrors AddHelperRules: all of U+0000-U+001F must stay escaped
+    // per RFC 8259 section 7.
+    int32_t unescaped_character =
+        builder_.AddCharacterClass({{0, 0x1f}, {'"', '"'}, {'\\', '\\'}}, true);
+    int32_t short_escape_character = builder_.AddCharacterClass(
+        {{'"', '"'},
+         {'\\', '\\'},
+         {'/', '/'},
+         {'b', 'b'},
+         {'f', 'f'},
+         {'n', 'n'},
+         {'r', 'r'},
+         {'t', 't'}}
+    );
+    int32_t hex_character = builder_.AddCharacterClass({{'0', '9'}, {'A', 'F'}, {'a', 'f'}});
+    int32_t non_surrogate_lead =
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Re-checked this against today's `main` (`1de4247`) before nudging, since the file moved under me.

The branch this PR changes is unchanged, now at `cpp/json_schema_converter.cc:2391-2395`:

```cpp
  // Check for length constraints
  if (spec.min_length != 0 || spec.max_length != -1) {
    int32_t character =
        builder_.AddCharacterClass({{'"', '"'}, {'\\', '\\'}, {'\r', '\r'}, {'\n', '\n'}}, true);
    int32_t body = Repeat(rule_name + "_characters", character, spec.min_length, spec.max_length);
```

so a length-bounded string still has no escape alternative: `\"`, `\\`, `\/`, `\b`, `\f`, `\n`, `\r`,
`\t` and `\uXXXX` are unemittable, while raw control characters are admitted. Five lines below, the
unbounded path returns `RuleRef(kBasicStringSub)`, and that helper rule is built from
`escaped_character | unicode_escape` (`:1877-:1895`) — same function, two string grammars, only one of
which can spell an escape. That asymmetry is the whole claim of this PR; `maxLength` is what routes a
schema onto the branch that lost it.

What I verified and what I did not:

- Verified by reading `origin/main` after `git fetch`: the two lines my diff removes are byte-identical
  to `:2393` today, and the API reports `mergeable: true / mergeable_state: clean`, so nothing here
  needs a rebase before review. My hunk header says `2370` because the file grew by ~22 lines above
  that point since I branched — line drift only, no textual conflict.
- Not verified: I read the source, I did not rebuild `libxgrammar` on this base, so the test evidence in
  the description is from `78ed389` as originally written rather than a re-run against today's `main`.

If the surrogate-pair enumeration in the current diff is more surface area than you want in
`GenerateString`, the same failure is covered by a smaller change: the eight basic escapes plus one
plain `\uXXXX` branch, and a follow-up for the astral-plane counting. Say which shape you would merge
and I will re-cut this to it.

PR #901, opened 2026-09-17
feat: support string exclusions in JSON schema structural tags

izhuhaoran:
Kimi-K3 can emit structural control markers inside a tool argument's string value. The schema grammar accepts them as text, while the tool-call parser treats them as structure.

This is the original detokenized badcase with prose redacted, the tool renamed to `reply`, and line breaks added for readability. The control-marker order, argument names, and repeated `index="1"` are preserved:

```text
<|open|>think<|sep|>[reasoning redacted]<|close|>think<|sep|>
<|open|>response<|sep|>[intro redacted]<|close|>response<|sep|>
<|open|>tools<|sep|>
<|open|>call tool="reply" index="1"<|sep|>
<|open|>argument key="response" type="string"<|sep|>[answer redacted]
<|close|>response<|sep|>
<|open|>tools<|sep|>
<|open|>call tool="reply" index="1"<|sep|>
<|open|>argument key="response" type="string"<|sep|>[follow-up redacted]<|close|>argument<|sep|>
<|open|>argument key="task_progress" type="string"<|sep|>[progress redacted]<|close|>argument<|sep|>
<|close|>call<|sep|>
<|close|>tools<|sep|>
<|close|>message<|sep|>[EOS]
```

Immediately after `[answer redacted]`, the first `response` argument is still open. The model emits `<|close|>response<|sep|>` where it should complete `<|close|>argument<|sep|>`, then starts another tools/call block before closing the original one. This is not a missing JSON quote: K3 string arguments are raw text. Without exclusions, the grammar can swallow the unexpected markers and repeated call/argument opening as part of the first string, until the later `<|close|>argument<|sep|>`.

Add optional `JSONSchemaFormat.excludes` to reserve substrings within string values and property names, including nested JSON:

```python
JSONSchemaFormat(
    json_schema={
        "type": "object",
        "properties": {"message": {"type": "string"}},
        "required": ["message"],
    },
    style="kimi_k3_xml",
    excludes=["<|open|>", "<|close|>", "<|sep|>"],
)
```

This is a structural-tag format option, not a JSON Schema keyword. It matches emitted text before JSON/XML unescaping, including matches split across tokens. It does not filter token IDs or join separate strings. An omitted or empty list applies no substring filter.

Existing string constraints remain in effect. JSON delimiter quotes and XML wrappers are outside the exclusion scope: `<|close|>` can still begin a valid `<|close|>argument<|sep|>` terminator, but cannot continue with `response` as string content. Tokens spanning a value's end and its enclosing suffix remain valid.

The Kimi-K3 builtin now supplies these three exclusions for every tool-argument schema in auto, required, and forced tool modes. The existing `exclude_special_tokens=False` option disables this filter. Free-text reasoning/response exclusions retain their current scope.

Compiler changes preserve existing string constraints and fix Unicode range/empty-language handling needed to prevent excluded branches from accepting multibyte or overlong UTF-8 input.

Non-empty exclusions currently reject `cohere_xml` and `minimax_m3_xml`; XML styles also reject exclusions starting or ending with formatting whitespace (space, tab, CR, LF). Compilation fails explicitly if constraints cannot be lowered safely or exceed construction budgets.

Validation:

- JSON-schema, structural-tag, Unicode, serialization, and native regression tests; repository formatting/lint checks and TypeScript type checking.
- K3 builtin tests cover default exclusions and opt-out, all three tool-choice modes, raw and nested JSON strings, and complete valid argument/call endings.
- Token-mask replay rejects the original bad output at `response`, while accepting both the repaired output and the subsequent valid model response. A synthetic vocabulary also verifies split markers and tokens crossing valid end boundaries; mask and `accept_token` results agree.


Seven-Streams:
@izhuhaoran Thanks for your contributions! I'll split the pr for further review.

izhuhaoran:
> @izhuhaoran Thanks for your contributions! I'll split the pr for further review.

@Seven-Streams  Thanks for helping split this up! Let me know if any changes are needed on my side.

